PR #19, opened 2016-12-07
fix bug in configuration of maxiomum rate

CodingCat:


src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
```diff
@@ -91,7 +91,7 @@ class EventHubsClientWrapper extends Serializable {
       MAXIMUM_PREFETCH_COUNT = MINIMUM_PREFETCH_COUNT
     } else if (maximumEventRate >= MINIMUM_PREFETCH_COUNT &&
       maximumEventRate < MAXIMUM_PREFETCH_COUNT) {
-      MAXIMUM_PREFETCH_COUNT = MAXIMUM_EVENT_RATE + 1
```
CodingCat: it shall be maximum event rate, otherwise, maximum_prefetch_count will be 1, violating the rule of 10 - 999

PR #81, opened 2017-05-05
EnqueueTime issue with first job fix

kpratik16:
Instead of relying on timeout of eventhub client, using Last Sequence no property of RunTimeInfo to determine if eventhub contains more events.

### DIFF
--- a/core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
+++ b/core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
@@ -19,10 +19,8 @@ package org.apache.spark.streaming.eventhubs
 
 // scalastyle:off
 import scala.collection.mutable.ListBuffer
-
 import com.microsoft.azure.eventhubs.EventData
 import org.apache.hadoop.conf.Configuration
-
 import org.apache.spark.{Partition, SparkContext, TaskContext}
 import org.apache.spark.annotation.DeveloperApi
 import org.apache.spark.rdd.RDD
@@ -60,10 +58,9 @@ class EventHubRDD(
     }.toArray
   }
 
-  private def wrappingReceive(
-      eventHubNameAndPartition: EventHubNameAndPartition,
-      eventHubClient: EventHubsClientWrapper,
-      expectedEventNumber: Int): List[EventData] = {
+  private def wrappingReceive(eventHubNameAndPartition: EventHubNameAndPartition,
+                              eventHubClient: EventHubsClientWrapper, expectedEventNumber: Int):
+  List[EventData] = {
     val receivedBuffer = new ListBuffer[EventData]
     val receivingTrace = new ListBuffer[Long]
     var cnt = 0
@@ -72,16 +69,21 @@ class EventHubRDD(
         throw new Exception(s"$eventHubNameAndPartition cannot return data, the trace is" +
           s" ${receivingTrace.toList}")
       }
-      val receivedEventsItr = eventHubClient.receive(expectedEventNumber - receivedBuffer.size)
-      if (receivedEventsItr == null) {
-        // no more messages
-        return receivedBuffer.toList
-      }
-      val receivedEvents = receivedEventsItr.toList
+      val receivedEvents = eventHubClient.receive(expectedEventNumber - receivedBuffer.size).
+        toList
       receivingTrace += receivedEvents.length
       cnt += 1
       receivedBuffer ++= receivedEvents
+
+      receivingTrace += receivedEvents.length
+      if(receivedEvents.length > 0 && eventHubClient.partitionRuntimeInformation != null) {
+        if (receivedEvents.last.getSystemProperties.getSequenceNumber ==
+          eventHubClient.partitionRuntimeInformation.getLastSequenceNumber) {
+          return receivedBuffer.toList
+        }
+      }
     }
+
     receivedBuffer.toList
   }
 
--- a/core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
+++ b/core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
@@ -37,6 +37,7 @@ private[eventhubs] class EventHubsClientWrapper extends Serializable with EventH
   with Logging {
 
   var eventhubsClient: AzureEventHubClient = _
+  var partitionRuntimeInformation: ReceiverRuntimeInformation = _
 
   // TODO: the design of this class is not simple enough
   // ideally, we shall not require the user to explicitly call createReceiver first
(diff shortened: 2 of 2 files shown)


core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -19,10 +19,8 @@ package org.apache.spark.streaming.eventhubs
 
 // scalastyle:off
 import scala.collection.mutable.ListBuffer
-
```
CodingCat: recover the changes

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -19,10 +19,8 @@ package org.apache.spark.streaming.eventhubs
...
 import com.microsoft.azure.eventhubs.EventData
 import org.apache.hadoop.conf.Configuration
-
```
CodingCat: recover the changes

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -60,10 +58,9 @@ class EventHubRDD(
     }.toArray
   }
 
-  private def wrappingReceive(
-      eventHubNameAndPartition: EventHubNameAndPartition,
-      eventHubClient: EventHubsClientWrapper,
-      expectedEventNumber: Int): List[EventData] = {
+  private def wrappingReceive(eventHubNameAndPartition: EventHubNameAndPartition,
+                              eventHubClient: EventHubsClientWrapper, expectedEventNumber: Int):
```
CodingCat: recover the changes

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -60,10 +58,9 @@ class EventHubRDD(
...
+  List[EventData] = {
```
CodingCat: recover the changes

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -72,16 +69,21 @@ class EventHubRDD(
...
-      }
-      val receivedEvents = receivedEventsItr.toList
+      val receivedEvents = eventHubClient.receive(expectedEventNumber - receivedBuffer.size).
+        toList
       receivingTrace += receivedEvents.length
       cnt += 1
       receivedBuffer ++= receivedEvents
+
+      receivingTrace += receivedEvents.length
+      if(receivedEvents.length > 0 && eventHubClient.partitionRuntimeInformation != null) {
```
CodingCat: space between if and (

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -72,16 +69,21 @@ class EventHubRDD(
...
+        if (receivedEvents.last.getSystemProperties.getSequenceNumber ==
+          eventHubClient.partitionRuntimeInformation.getLastSequenceNumber) {
```
CodingCat: you need to keep `receivedEvents == null` to ensure that we have something old to prevent regression (since we have no control on EventHubs server behavior

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubRDD.scala
```diff
@@ -72,16 +69,21 @@ class EventHubRDD(
...
+          return receivedBuffer.toList
+        }
+      }
     }
+
```
CodingCat: recover the chagnes

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
```diff
@@ -131,22 +132,27 @@ private[eventhubs] class EventHubsClientWrapper extends Serializable with EventH
     // Create Eventhubs client
     eventhubsClient = EventHubClient.createFromConnectionStringSync(connectionString)
 
+    // Enabling flag to receive runtime info from eventhub client.
+    val receiverOptions = new ReceiverOptions
+    receiverOptions.setReceiverRuntimeMetricEnabled(true)
```
CodingCat: what's the cost of this? we only need this in the first batch, but whey we want to enable it for every batch?

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
```diff
@@ -177,7 +183,13 @@ private[eventhubs] class EventHubsClientWrapper extends Serializable with EventH
   def receive(expectedEventNum: Int): Iterable[EventData] = {
     val events = eventhubsReceiver.receive(
       math.min(expectedEventNum, eventhubsReceiver.getPrefetchCount)).get()
-    if (events == null) Iterable.empty else events.asScala
+    if (events == null) {
+      Iterable.empty
+    }
+    else {
```
CodingCat: put it to the above line

core/src/main/scala/org/apache/spark/streaming/eventhubs/EventHubsClientWrapper.scala
CodingCat: What if the customer only have 32 bytes payload...how many bytes we received once we turned on for every batch?

kpratik16: I will check that with Sreeram.

CodingCat: You can calculate by yourself, class header+all fields in runtimeInfo class...so I am thinking if we really want to merge this fix..once their 0.14 is out we can get start seq easily....and all of these can be removed

kpratik16: Its approx about (26 bytes + 2 Instant java object). Not sure about size of Instant object.

It definitely is a bit expensive operation. Do you have any approximate ETA for this feature using start seq number?

CodingCat: https://github.com/Azure/azure-event-hubs-java/blob/b378b50b25253c036dad62cffbebe490f9bcfe15/azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ReceiverRuntimeInformation.java#L17, the last sequence number will be 64 bytes....so I think it is too expensive to introduce as a temporal solution

I do not know their plan, maybe soon, since they mentioned this feature to me several weeks ago

CodingCat:
I will close this one since the issue will be resolve by https://github.com/hdinsight/spark-eventhubs/issues/76 which is coming soon

PR #123, opened 2017-08-17
check for null reference in RDD to DataFrame conversion

### LINKED ISSUE #121: getProperties failed with NullPointerException, if null present in properties values during map generation
Caused by: java.lang.NullPointerException
	at org.apache.spark.sql.streaming.eventhubs.EventHubsSource$$anonfun$7$$anonfun$apply$5.apply(EventHubsSource.scala:280)

Please, fix v.toString

sabeegrewal:
close #121 

If the value is null, I place an empty string in its place. Following the same convention set up on line 277 within the same method.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/streaming/eventhubs/EventHubsSource.scala
+++ b/core/src/main/scala/org/apache/spark/sql/streaming/eventhubs/EventHubsSource.scala
@@ -278,7 +278,7 @@ private[spark] class EventHubsSource(
             })
           } else {
             Seq(eventData.getProperties.asScala.map { case (k, v) =>
-              k -> (if (v == null) "" else v.toString) })
+              k -> (if (v == null) null else v.toString) })
           }
         } else {
           Seq()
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/apache/spark/sql/streaming/eventhubs/EventHubsSource.scala
```diff
@@ -277,7 +277,8 @@ private[spark] class EventHubsSource(
               eventData.getProperties.asScala.getOrElse(k, "").toString
             })
           } else {
-            Seq(eventData.getProperties.asScala.map { case (k, v) => k -> v.toString })
+            Seq(eventData.getProperties.asScala.map { case (k, v) =>
+              k -> (if (v == null) "" else v.toString) })
```
CodingCat: would you please add a unit test and also try if Spark can serialize null value (I think it can ), if Spark can serialize it, we do not even need to do any translation

sabeegrewal:
added the unit test, everything looks good to me. Let me know if you think anything else should be added!

CodingCat:
others looks good to me, would you please pull the latest master to trigger the test again?

ogidogi:
Can we expect this fix will be in version 2.1.3

CodingCat:
yes

PR #135, opened 2017-09-07
updating receiver identifier with different info

### LINKED ISSUE #118: More readable identifier id warning for Eventhubs recriver
To upgrade `azure-event-hubs-java` version to `0.14.3`, code set identifier id in [commit-109](https://github.com/hdinsight/spark-eventhubs/commit/e9babda4a077d81160280ad37793f46bb93c9bfb) as format: `$consumerGroup-$eventHubsName-$partitionId-$currentOffset`

However, this will cause receiver failure due to validation in [azure-event-hubs code](https://github.com/Azure/azure-event-hubs-java/blob/0.14.3/azure-eventhubs/src/main/java/com/microsoft/azure/eventhubs/ReceiverOptions.java#L73), which will throw exception like "receiverIdentifier length cannot exceed 64". Actually, this exception message is not so readable, and users might need some time to find the root cause, I think it's better to throw exception in Spark-eventhubs' [EventHubsClientWrapper.java](https://github.com/hdinsight/spark-eventhubs/blob/master/core/src/main/scala/org/apache/spark/eventhubscommon/client/EventHubsClientWrapper.scala#L152) file to tell users to shorten `consumerGroup` or `eventHubsName` names.

sabeegrewal:
close #118 

consumerGroup, eventHubsName, and partitionId already exist in the receiver state without adding them to the receiverIdentifier. Adding spark executorId and spark TaskId, so receivers can be matched with specific tasks and machines.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubscommon/client/EventHubsClientWrapper.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubscommon/client/EventHubsClientWrapper.scala
@@ -19,9 +19,11 @@ package org.apache.spark.eventhubscommon.client
 import java.time.Instant
 
 import scala.collection.JavaConverters._
+
 import com.microsoft.azure.eventhubs.{EventHubClient => AzureEventHubClient, _}
 import com.microsoft.azure.servicebus._
 import EventHubsOffsetTypes.EventHubsOffsetType
+
 import org.apache.spark.{SparkEnv, TaskContext}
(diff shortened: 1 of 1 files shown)


CodingCat:
a stylistic issue, would you please file the PR for 2.0.x branch as well?

CodingCat:
BTW if all information has been in eventhubs side, do we really need to call this API here? can we just simply remove this?

PR #138, opened 2017-09-07
[spark-2.0.x] updating receiverOptions with information related to spark application

sabeegrewal:


CodingCat:
not sure why it does not work here, you can change level to ERROR directly....

PR #140, opened 2017-09-14
added more tracing for eventhub api calls

sabeegrewal:


core/src/main/scala/org/apache/spark/eventhubscommon/client/AMQPEventHubsClient.scala
```diff
@@ -44,6 +44,7 @@ private[client] class AMQPEventHubsClient(
       targetEventHubNameAndPartitions: List[EventHubNameAndPartition]) = {
     val results = new mutable.HashMap[EventHubNameAndPartition, EventHubPartitionRuntimeInformation]
     try {
+      logInfo("Getting PartitionRuntimeInformation from EventHubs instance.")
```
CodingCat: why it is necessary? the latency from something like `JobScheduler's start a streaming job `  to `highest offset, xxxx` is approximately the latency between these two lines

core/src/main/scala/org/apache/spark/eventhubscommon/rdd/EventHubsRDD.scala
```diff
@@ -82,6 +84,7 @@ private[spark] class EventHubsRDD(
       cnt += 1
       receivedBuffer ++= receivedEvents
     }
+    logInfo(s"Finished ingesting events from partition ${eventHubNameAndPartition.partitionId}")
```
CodingCat: Why it is necessary, we have https://github.com/sabeegrewal/spark-eventhubs/blob/9e497a4f7d5382fcdbc33a6ba1e409581c088996/core/src/main/scala/org/apache/spark/eventhubscommon/rdd/EventHubsRDD.scala#L120

PR #199, opened 2017-10-31
Add support for Azure German Cloud (and China, US govt, etc)

p2bauer:
We're blocked deploying our EH spark-based solution to a customer site due to not being able to run against an eventhub located in the german cloud.

I'm having issues building the solution locally with maven so I'm not able to test properly - but I thought I'd open the PR now so that maybe reviewers can get eyes-on as soon as possible.

Essentially this just allows for `eventhubs.namespace` to either be the namespace name, or the fully qualified namespace URL - and uses the existing ConnectionStringBuilder overload in the java lib.

Once tested and merged, how would we go about getting an updated build published to maven so we can proceed with deployment?

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubscommon/client/EventHubsClientWrapper.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubscommon/client/EventHubsClientWrapper.scala
@@ -120,7 +120,7 @@ private[spark] class EventHubsClientWrapper(
   private[spark] def createReceiverInternal(partitionId: String,
                                             offsetType: EventHubsOffsetType,
                                             currentOffset: String): Unit = {
-    eventhubsClient = EventHubClient.createFromConnectionStringSync(connectionString)
+    eventhubsClient = EventHubClient.createFromConnectionStringSync(connectionString.toString)
 
     eventhubsReceiver = offsetType match {
       case EventHubsOffsetTypes.None | EventHubsOffsetTypes.PreviousCheckpoint |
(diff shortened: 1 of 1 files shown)


p2bauer:
investigating build failures now :-)

sabeegrewal:
Hey Paul, thanks for the PR!! I wanna make sure I understand your change correctly. Instead of providing the namespace, name, policy name, and key separately, you'd like the option to provide the entire connection string in the namespace name?

Let me know if I'm understanding that correctly, and we can discuss the best way to handle it 👍

sabeegrewal:
Also, is a staging build sufficient for now or do you guys need a release in maven central? The staging build can be made available much quicker, so you guys can be unblocked. We're happy to release to maven central to accommodate you guys, but that would take more time...probably end of this/early next week.

p2bauer:
@sabeegrewal - i'm totally open to an alternative solution on how to handle EH configuration (and we can just discard this PR) - this was just the most minimal change I could come up with.

Re: maven, staging build would be great as well, just let me know.

p2bauer:
oh, and I don't necessarily need the whole connection string - just the fully qualified namespace url instead of namespace name.

ie. I could specify `<NAMESPACE>.servicebus.windows.net` instead of just `<NAMESPACE>` - since in the case of the german cloud, the suffix would not be .servicebus.windows.net.

sabeegrewal:
Ok, let's do a staging version for now. I'll expose the ability to add an EventHubs URI to the user, and you can set your URI there. Could you close this out and open an issue? We can continue this conversation there :)

PR #201, opened 2017-11-01
Moving to Spark 2.2 and adding support for non-public clouds

### LINKED ISSUE #200: Add ability to specify full eventhub namespace url (versus just namespace name)
This will let us use to stream from Event Hubs located in non-public Azure Clouds, such as the German or Chinese cloud.

sabeegrewal:
- Adopting spark 2.2
- Allowing URI or namespace (for use with private clouds...German, US Gov't, Chinese, etc)
- Temporarily disabling struct stream tests

close #200

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -23,7 +23,7 @@
   <parent>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>spark-streaming-eventhubs_connector_2.11</artifactId>
-    <version>2.1.6-SNAPSHOT</version>
+    <version>2.2.0-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
   <artifactId>spark-streaming-eventhubs_2.11</artifactId>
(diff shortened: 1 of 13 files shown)


p2bauer:
@sabeegrewal i noticed that this was merged into the spark 2.2.x branch?  We're running 2.1.x in all our environments (both in german and public cloud), and I'm worried this update could break our checkpoint/progress files.  Is this completely backwards compatible with deployments using your 2.1.x libraries for direct streaming?

sabeegrewal:
Hey, good point....I'll push a 2.1.6-SNAPSHOT for you as well. Just a sec!

sabeegrewal:
Hey Paul, just pushed a 2.1.6-SNAPSHOT with a Spark 2.1 dependency. 👍

slyons:
Seeing as how Spark 2.2 is not officially part of an HDI release yet (only 2.0 and 2.1 are available) it might be prudent to maintain releases for each of those versions.

sabeegrewal:
Hey Scott, for releases we're going to maintain versioning like we've been doing (separate versions for Spark 2.0, 2.1, 2.2, etc). I just got a little ahead of myself on pushing staging versions cause I'm trying to support Spark 2.2 (and the current 2.3 SNAPSHOTS). Having 2.2 and 2.1 dev work in a single 'dev' branch could be improved on, and we'll be addressing that soon! 

Thanks for the feedback :)

PR #212, opened 2017-11-14
[WIP] User configuration re-write

sabeegrewal:
This pull request introduces the EventHubsConf. EventHubsConf will be the main way users configure the EventHubs. Introducing this abstraction:
  - makes it easier and less error prone for users to configure their application
  - Provides more flexibility for us to avoid breaking changes
  - Allows users to change receiver and operation timeouts
  - Allows users to configure max rate, start offsets, and start enqueue times on a per partition basis
  - Allows users to start from the EndOfStream or StartOfStream

 Example usage:
```scala
val ehConf = EventHubsConf()
  .setNamespace("namespace")
  .setName("name")
  .setKeyName("keyName")
  .setKey("key")
  .setPartitionCount("8")
  .setStartOffsets(0 to 3, 5000)
  .setStartOffsets(4 to 7, 1000) 
```

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/common/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/common/EventHubsConf.scala
@@ -1,0 +1,190 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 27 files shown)


slyons:
Instead of a final class, could the config class be made as a trait instead? This would allow property enforcement while enabling loading from different sources.

Issue #202, opened 2017-11-06
Add PySpark support


mattdot:
### Feature Request
Please add support for PySpark.  Developers using the Python API would also like to receive events from Azure Event Hubs!
If/When you do add support, please also add an example to the docs

sabeegrewal:
Hey Matt, we're currently developing an EventHubs client in Python. When that's done, we'll definitely look into supporting PySpark. I can't give a strong timeline at this point, but I'll post updates here! Thanks for the request 👍

mmaitre314:
This seems to be almost working in Azure Databricks PySpark notebooks (minus Issue #213). In a regular PySpark app something similar should work after loading the EventHub JAR:

```
from pyspark.sql import SparkSession

spark = (SparkSession.builder
    .appName("TestApp")
    .config("spark.jars", "/<snip>/spark-streaming-eventhubs_2.11-2.1.5.jar")
    .getOrCreate())
```

One thing I am not clear about is how to convert the event body from bytes to string. In Java/Scala that can be done through a UDF. PySpark also supports UDFs, but with a serialization overhead.

sabeegrewal:
Hey Matthieu, check the other issue you opened (I've recently replied!), and see if that works for you. Keep me posted in #213 and then we can deal with the event body conversion issues :)

happy thanksgiving!

romitgirdhar:
I tested this using PySpark3 in HDInsight and it seems to be working now. Happy to contribute the code as a sample.

sabeegrewal:
Sweet! That would be great....a PR or some quick documentation on how you're using the Scala libraries with PySpark. Either one works!

Also, in regards to Matthieu's issue, there seems to be casting issues that show up in Spark 2.2 and beyond (which I'm currently looking into).

sabeegrewal:
Okay, a new version of `azure-eventhubs-databricks_2.11` has been released and should be available soon. @mmaitre314 let me know if you have issues with `v3.4.0`!

sabeegrewal:
I have a customer that's had some success using PySpark notebooks in Databricks as well. I'll get some documentation up on that and then close this issue out. For the next 6 months or so, that'll be the only PySpark support we work on.

PR #215, opened 2017-11-30
Sample for using the eventhubs-spark JAR in Jupyter notebook using PySpark3

romitgirdhar:
This should also address issue #202


sabeegrewal:
Could you move the `.md` file to `/docs/PySpark/`? Instead of `examples`? 

LGTM tho, thank you so much for helping with this :)

romitgirdhar:
Done!

PR #218, opened 2017-12-08
[WIP] Library re-write (For Spark 2.2)

sabeegrewal:
Update (2/2)
We're tuning the connector with Spark 2.3. We'll bring fixes into this branch (Spark 2.2) before release. You can follow Spark 2.3 connector work on #229 

Update (2/1)
Everything is wrapped up, all tests are passing. I'm moving to documentation, examples, and stress testing. Thanks for your patience! 

---------------------
Re-write is mostly done, just writing tests now. A key focus for testing is making sure we leverage Spark's test utilities as much as possible so that we don't have to make many changes across versions. In other words, for tests we're going to reduce our dependency on Spark internals as much as we can. 

Next steps:
- ~~Complete Sequence Number support (ehConf should accept sequence number as a starting type)~~
- ~~RDD tests~~
- ~~DStream tests~~
- ~~Source tests~~
- ~~EventHubClient Tests~~

This wraps up the ongoing re-write of the library : #174 #182 #183 #184 #185 #188 #189 #190 #191 #192 #194 #196. When the changes are complete, I will update this with more specific info. Many of the changes are already listed on the wiki.


sabeegrewal:
from @slyons in #212 :
"Instead of a final class, could the config class be made as a trait instead? This would allow property enforcement while enabling loading from different sources." 

^ still need to do that!

core/src/main/scala/org/apache/spark/eventhubs/common/client/EventHubsClientWrapper.scala
```diff
@@ -18,69 +18,61 @@
...
-  private val ehPolicy = ehParams("eventhubs.policykey").toString
-  private val consumerGroup =
-    ehParams.getOrElse("eventhubs.consumergroup", EventHubClient.DEFAULT_CONSUMER_GROUP_NAME)
   private val connectionString =
     Try {
-      new ConnectionStringBuilder(ehNamespace, ehName, ehPolicyName, ehPolicy).toString
+      new ConnectionStringBuilder(ehNamespace, ehName, ehPolicyName, ehPolicy)
     } getOrElse Try {
-      new ConnectionStringBuilder(new URI(ehNamespace), ehName, ehPolicyName, ehPolicy).toString
+      new ConnectionStringBuilder(new URI(ehNamespace), ehName, ehPolicyName, ehPolicy)
```
slyons: I was thinking about this earlier and would it make more sense to have an optional "suffix" config parameter? Rather than doubling execution and trying to parse the full namespace we could have a suffix that defaults to the usual EH value if not present.

core/src/main/scala/org/apache/spark/eventhubs/common/client/EventHubsClientWrapper.scala
```diff
@@ -132,10 +124,21 @@ private[spark] class EventHubsClientWrapper(private val ehParams: Map[String, St
...
    */
-  override def lastEnqueuedTime(nameAndPartition: NameAndPartition): Option[Long] = {
+  override def lastEnqueuedTime(nameAndPartition: NameAndPartition): EnqueueTime = {
     try {
       val runtimeInfo = getRunTimeInfo(nameAndPartition)
-      Some(runtimeInfo.getLastEnqueuedTimeUtc.getEpochSecond)
+      runtimeInfo.getLastEnqueuedTimeUtc.getEpochSecond
+    } catch {
+      case e: Exception =>
+        e.printStackTrace()
```
slyons: Is printing the stack trace at this point needed? Surely it would bubble up

core/src/main/scala/org/apache/spark/eventhubs/common/client/EventHubsClientWrapper.scala
```diff
@@ -145,17 +148,76 @@ private[spark] class EventHubsClientWrapper(private val ehParams: Map[String, St
...
+    val result = new ConcurrentHashMap[PartitionId, SequenceNumber]()
+
+    if (startingWith equals "SequenceNumbers") {
+      (for (partitionId <- 0 until partitionCount)
+        yield
+          partitionId -> ehConf.startSequenceNumbers.getOrElse(partitionId,
+                                                               DefaultStartSequenceNumber)).toMap
+    } else {
+      val threads = for (partitionId <- 0 until partitionCount)
+        yield
```
slyons: All this could possibly be simplified using `.par.map`

core/src/main/scala/org/apache/spark/eventhubs/common/rdd/EventHubsRDD.scala
```diff
@@ -17,139 +17,116 @@
...
+  override def count: Long = offsetRanges.map(_.count).sum
+
+  override def isEmpty(): Boolean = count == 0L
+
+  override def take(num: Int): Array[EventData] = {
+    val nonEmptyPartitions =
+      this.partitions.map(_.asInstanceOf[EventHubsRDDPartition]).filter(_.count > 0)
+
+    if (num < 1 || nonEmptyPartitions.isEmpty) {
+      return new Array[EventData](0)
```
slyons: `Array.empty[EventData]` will skip an allocation

core/src/main/scala/org/apache/spark/sql/streaming/eventhubs/EventHubsSourceProvider.scala
```diff
@@ -43,50 +46,56 @@ private[sql] class EventHubsSourceProvider
...
+  private val FailOnDataLossOptionKey = "eventhubs.failondataloss"
 
-  private[eventhubs] def ifContainsPropertiesAndUserDefinedKeys(
-      parameters: Map[String, String]): (Boolean, Seq[String]) = {
+  def sourceSchema(parameters: Map[String, String]): StructType = {
     val containsProperties =
       parameters.getOrElse("eventhubs.sql.containsProperties", "false").toBoolean
+
     val userDefinedKeys = {
       if (parameters.contains("eventhubs.sql.userDefinedKeys")) {
```
slyons: Is there plans to open up typing for custom properties?

core/src/main/scala/org/apache/spark/eventhubs/common/client/EventHubsClientWrapper.scala
sabeegrewal: I like this idea! I'll get to it after tests are finished.

core/src/main/scala/org/apache/spark/eventhubs/common/rdd/EventHubsRDD.scala
sabeegrewal: Agreed, will change to ```return Array()```.

core/src/main/scala/org/apache/spark/sql/streaming/eventhubs/EventHubsSourceProvider.scala
sabeegrewal: Not explicitly. Once this re-write is complete, we'll start integration testing and planning nice things to layer on top of the new library. I will definitely contact you for input around then!

core/src/main/scala/org/apache/spark/eventhubs/common/EventHubsConf.scala
```diff
@@ -0,0 +1,506 @@
...
+import java.time.Duration
+import java.util.concurrent.ConcurrentHashMap
+
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.catalyst.util.CaseInsensitiveMap
+
+import scala.collection.JavaConverters._
+import language.implicitConversions
+
+// TODO: Can we catch malformed configs at compile-time?
```
sabeegrewal: Two changes coming to EventHubsConf:
1) An EventHubsConf will have to be constructed with a connection string. The setters for namespace, name, key name, and key will be removed. We expect most people to copy/paste their connection string from the service portal. For those that prefer to do it piecemeal (like I personally do), I'll expose the connection string builder from the new java client which users can use to make their connection string. This makes the configuration process simpler which makes the rest of the code cleaner and less bug-prone. 

2) The new EventHubs Java client will have a new EventPosition abstraction. I'll be removing the setters (and all internals) for seq number, offset, and enqueued time. I'll be replacing that with `setStartingPosition(EventPosition)` and `setStartingPosition(Map[PartitionId -> EventPosition])`. This also will clean things up quite a bit.

core/src/main/scala/org/apache/spark/eventhubs/common/package.scala
```diff
@@ -0,0 +1,70 @@
...
+
+package org.apache.spark.eventhubs
+
+import java.time.Duration
+
+import com.microsoft.azure.eventhubs.{ EventHubClient, PartitionReceiver }
+
+package object common {
+  val DefaultDomainName: String = "servicebus.windows.net"
+  val DefaultEnqueueTime: EnqueueTime = Long.MinValue
```
sabeegrewal: Bug -> this should be the min epoch time.

core/src/main/scala/org/apache/spark/eventhubs/common/package.scala
```diff
@@ -0,0 +1,70 @@
...
+  val StartOfStream = PartitionReceiver.START_OF_STREAM
+  val EndOfStream = PartitionReceiver.END_OF_STREAM
+  val DefaultMaxRatePerPartition: Rate = 10000
+  val DefaultStartOffset: Offset = -1L
+  val DefaultStartSequenceNumber: SequenceNumber = 0L
+  val DefaultReceiverTimeout: Duration = Duration.ofSeconds(5)
```
sabeegrewal: We'll make this 60 seconds

core/src/main/scala/org/apache/spark/eventhubs/common/package.scala
```diff
@@ -0,0 +1,70 @@
...
+import java.time.Duration
+
+import com.microsoft.azure.eventhubs.{ EventHubClient, PartitionReceiver }
+
+package object common {
+  val DefaultDomainName: String = "servicebus.windows.net"
+  val DefaultEnqueueTime: EnqueueTime = Long.MinValue
+  val StartOfStream = PartitionReceiver.START_OF_STREAM
+  val EndOfStream = PartitionReceiver.END_OF_STREAM
+  val DefaultMaxRatePerPartition: Rate = 10000
```
sabeegrewal: Drop this to 1000

core/src/main/scala/org/apache/spark/eventhubs/common/EventHubsUtils.scala
```diff
@@ -43,39 +38,17 @@ object EventHubsUtils {
...
-  def createDirectStream(ssc: StreamingContext,
-                         progressDir: String,
-                         ehParams: Map[String, String]): EventHubDirectDStream = {
-    val ehName = ehParams("eventhubs.name")
-    createDirectStreams(ssc, progressDir, Map(ehName -> ehParams))
+  def createRDD(sc: SparkContext,
+                ehConf: EventHubsConf,
+                offsetRanges: Array[OffsetRange]): EventHubsRDD = {
+    new EventHubsRDD(sc, ehConf, offsetRanges, EventHubsClientWrapper.apply)
   }
```
sabeegrewal: Add support for Java RDDs and Java DStreams

sabeegrewal:
Note, build will fail until the 1.0.0 of the java client is available on maven.

core/src/main/scala/org/apache/spark/eventhubs/package.scala
```diff
@@ -0,0 +1,66 @@
...
+import java.time.Duration
+import java.util.concurrent.Executor
+
+import com.microsoft.azure.eventhubs.EventHubClient
+import org.apache.spark.eventhubs.utils.Position
+
+package object eventhubs {
+
+  val DefaultEventPosition: Position = Position.fromStartOfStream()
+  val DefaultMaxRatePerPartition: Rate = 10000
```
sabeegrewal: Drop this to 1000

core/src/main/scala/org/apache/spark/eventhubs/EventHubsUtils.scala
```diff
@@ -0,0 +1,54 @@
...
+  def createDirectStream(ssc: StreamingContext, ehConf: EventHubsConf): EventHubsDirectDStream = {
+    new EventHubsDirectDStream(ssc, ehConf, EventHubsClientWrapper.apply)
+  }
+
+  def createRDD(sc: SparkContext,
+                ehConf: EventHubsConf,
+                offsetRanges: Array[OffsetRange]): EventHubsRDD = {
+    new EventHubsRDD(sc, ehConf, offsetRanges, EventHubsClientWrapper.apply)
+  }
+}
```
sabeegrewal: add support for Java RDDs and DStreams

PR #223, opened 2018-01-21
reduce memory usage in EventHubsRDD

tilumi:
Introduce a iterator to avoid retrieve all the data in memory for next computation, it retrieve small portion of data whenever it needs.


sabeegrewal:
Marking this as closed because fixing the tests isn't worth the time investment. If anyone needs these changes before the re-launch in mid-Feb, then we can revisit this.

sabeegrewal:
By the way, thanks for taking the time to share these!!!

PR #224, opened 2018-01-21
fix min enqueue time ignored bug when consuming multiple hubs

tilumi:


sabeegrewal:
Marking this as closed because fixing the tests isn't worth the time investment. If anyone needs these changes before the re-launch in mid-Feb, then we can revisit this.

PR #228, opened 2018-02-02
Fix links

thebillkidy:
We need to provide the protocol, since else github will show github.com/.../www.ms.portal.azure.com which is of course wrong

### Thanks for contributing! We appreciate it :) 

#### Again, thanks for taking the time! Feel free to reach out with any questions or concerns!


sabeegrewal:
Ty :)

PR #238, opened 2018-03-02
Marking receivers/senders as null for GC

sabeegrewal:


jaceklaskowski:
I wonder how you caught the issue? What was the reason for the change? Any memory leaks? I'd appreciate some more information. Thanks.

sabeegrewal:
Hey Jacek! I've read a bit of your writing so it means a lot that you'd take time to look through the repo :) I'm looking forward to working more with you and others in Spark community. 

Anyways, we have long-running streams to stress test the integration. In a heap snapshot, we saw thousands of receiver instances for a single machine (we expect the receiver count to main quite low). So, we knew something was up with the receiver cleanup which resulted in this PR. Let me know if you'd like more detail 👍

PR #248, opened 2018-03-05
Key EventHubClientConnectionPool based on ConnectionString instead of EventHubName

SreeramGarlapati:
Current key for EHClient connection pool is based off of the EventHubName. 
Some scenarios might see trouble based on this:
1) using different `access keys` across 2 streams in the same spark job
2) if the access keys are rolled - cluster might need to restarted
3) if eventhubs names are same across 2 namespaces used.
etc..


core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -49,6 +49,7 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
       logInfo(s"Borrowing client. EventHub name: ${ehConf.name}")
     }
     logInfo(s"Available clients: {${pool.size}}. Total clients: ${count.get}")
+
```
sabeegrewal: Remove

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -59,51 +60,72 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
 
+/**
+  * Represents an object being pooled in {@link ClientConnectionPool}
+  * @param ehConf The Event Hubs configurations corresponding to the EventHubClient.
+  * @param ehClient The EventHubClient to be pooled.
+  */
+class ClientConnectionPoolObject(private[this] val ehConf: EventHubsConf, private[this] val ehClient: EventHubClient) {
+
+  def getEventHubsConf: EventHubsConf = {
+    return ehConf;
```
sabeegrewal: no `;`s and no `return`s in Scala

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -59,51 +60,72 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
+  }
+
+  def getEventHubClient: EventHubClient = {
+    return ehClient;
```
sabeegrewal: same here

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -59,51 +60,72 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
+  }
+}
+
 object ClientConnectionPool extends Logging {
 
   private def notInitializedMessage(name: String): String = {
-    s"Connection pool is not initialized for EventHubs: $name"
+    s"Connection pool is not initialized for EventHub: ${name}"
```
sabeegrewal: remove braces around `name`

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -59,51 +60,72 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
+    if (!isInitialized(hashKey)) {
+      val message = notInitializedMessage(ehConf.name)
       throw new IllegalStateException(message)
     }
   }
 
-  private def get(name: String): ClientConnectionPool = pools.synchronized {
-    pools.getOrElse(name, {
-      val message = notInitializedMessage(name)
+  private def get(ehConf: EventHubsConf): ClientConnectionPool = pools.synchronized {
```
sabeegrewal: Can we change the signature to `key: String`?

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -59,51 +60,72 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
+    val hashKey = ehConf.connectionString
+    pools.getOrElse(hashKey, {
+      val message = notInitializedMessage(ehConf.name)
```
sabeegrewal: We would have to pass the `namespace` + `name`, right?

sabeegrewal:
Going to close this for now until we can spend more time on it. In addition to the review comments, I don't think we need an `ehConf` in the connection pool object class and I think we need a better name than `ClientConnectionPoolObject`. 

We probably will need to do this eventually though - we'll revisit it soon.

Issue #249, opened 2018-03-06
EventDataSchema: Change Offset type from Long to String


SreeramGarlapati:
1) This will the spark adapter in sync with type of `EventData.offset` exposed by the java library.
2) Offset is an opaque object that EventHubs service constructs – to represent the position of EventData on the stream – and represented as a string – so that EventHubs service gets' to keep the flexibility to change - the type.

PR #253, opened 2018-03-06
Change offset column from Long Type to String Type

SreeramGarlapati:
Close: #249


SreeramGarlapati:
@sabeegrewal - the structured-streaming-eventhubs-integration.md - somehow already has the correct offset type already!

PR #270, opened 2018-03-09
Minor cleanup and documentation edits

sabeegrewal:
Some cleanup to improve the code and documentation quality. 

@jaceklaskowski, this should cover everything you mentioned expect for the `null`s in `EventPosition`. I'll address those in a different PR down the road! Thanks for the feedback :) Much appreciated.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -10,7 +10,7 @@ see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq
 
 To build and test this locally, make sure you've done the following:
 - Java 1.8 SDK is installed
-- Maven 3.x is installed (or SBT version 0.13.13)
+- [Maven 3.x](https://maven.apache.org/download.cgi) is installed (or [SBT version 1.x](https://www.scala-sbt.org/1.x/docs/index.html))
 - A supported version of Apache Spark is installed (see [Latest Releases](/README.md#latest-releases) for supported versions). 
 
 After that, cloning the code and running `mvn clean package` should successfully 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -8,11 +8,9 @@ see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq
 
 ## Getting Started
 
-This library is relatively easy to build. To build and test this locally, 
-make sure you've done the following:
+To build and test this locally, make sure you've done the following:
 - Java 1.8 SDK is installed
-- Maven 3.x is installed
-- Scala 2.11.X is installed
+- Maven 3.x is installed (or SBT version 0.13.13)
 - A supported version of Apache Spark is installed (see [Latest Releases](/README.md#latest-releases) for supported versions). 
 
 After that, cloning the code and running `mvn clean package` should successfully 
@@ -41,14 +39,14 @@ Then add the following dependency declaration:
     <dependency>
         <groupId>com.microsoft.azure</groupId>
         <artifactId>azure-eventhubs-spark_[2.XX]</artifactId>
-        <version>2.1.6-SNAPSHOT</version>
+        <version>2.3.0</version>
     </dependency>
 ```
 
 ### SBT Dependency
 
     // https://mvnrepository.com/artifact/com.microsoft.azure/azure-eventhubs-spark_2.11
-    libraryDependencies += "com.microsoft.azure" % "azure-eventhubs-spark_2.11" % "2.1.6-SNAPSHOT"
+    libraryDependencies += "com.microsoft.azure" %% "azure-eventhubs-spark" %% "2.3.0"
  
 ## Filing Issues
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/apache/spark/eventhubs/EventPosition.scala
```diff
@@ -69,7 +69,7 @@ object EventPosition {
    * @return An [[EventPosition]] instance.
    */
   def fromOffset(offset: String): EventPosition = {
-    EventPosition(offset = offset, isInclusive = true)
+    EventPosition(offset = offset)
```
jaceklaskowski: Could we remove the duplication?

core/src/main/scala/org/apache/spark/eventhubs/EventPosition.scala
```diff
@@ -81,7 +81,7 @@ object EventPosition {
    */
   def fromSequenceNumber(seqNo: SequenceNumber): EventPosition = {
     require(seqNo >= 0L, "Please pass a positive sequence number.")
-    EventPosition(seqNo = seqNo, isInclusive = true)
+    EventPosition(seqNo = seqNo)
```
jaceklaskowski: Could we remove the duplication?

docs/structured-streaming-eventhubs-integration.md
```diff
@@ -179,9 +179,7 @@ val df = spark
   .load()
   
 // Select body column as a String
```
jaceklaskowski: Remove the comment

core/src/main/scala/org/apache/spark/eventhubs/EventPosition.scala
sabeegrewal: We have to keep this duplication, everything else is cleaned up 👍

.github/CONTRIBUTING.md
```diff
@@ -8,11 +8,9 @@ see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq
 
 ## Getting Started
 
-This library is relatively easy to build. To build and test this locally, 
-make sure you've done the following:
+To build and test this locally, make sure you've done the following:
 - Java 1.8 SDK is installed
-- Maven 3.x is installed
-- Scala 2.11.X is installed
+- Maven 3.x is installed (or SBT version 0.13.13)
```
jaceklaskowski: I'd fix the version to "1.1.1" since this is the latest supported one. People could use earlier versions on their own, but the official docs (since we're at it) should promote the latest and greatest if it just works.

I also thought about linking the tools to their websites (perhaps even the download sections), but that could be just nitpicking.

README.md
```diff
@@ -25,7 +25,10 @@ By making Event Hubs and Spark easier to use together, we hope this connector ma
 |Spark Version|Package Name|Package Version|
 |-------------|------------|----------------|
 |Spark 2.3|azure-eventhubs-spark_2.11|[![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.microsoft.azure/azure-eventhubs-spark_2.11/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.microsoft.azure/azure-eventhubs-spark_2.11)|
-|Spark 2.2|azure-eventhubs-spark_2.11|[![Maven Central](https://img.shields.io/maven-central/v/com.microsoft.azure/azure-eventhubs-spark_2.11/2.2.0.svg)](https://maven-badges.herokuapp.com/maven-central/com.microsoft.azure/azure-eventhubs-spark_2.11)|
+|Spark 2.2|azure-eventhubs-spark_2.11|[![Maven Central](https://img.shields.io/maven-central/v/com.microsoft.azure/azure-eventhubs-spark_2.11/2.2.0-PREVIEW.svg)](https://maven-badges.herokuapp.com/maven-central/com.microsoft.azure/azure-eventhubs-spark_2.11)|
```
jaceklaskowski: Why do the two use `img.shields.io` while 2.3 `maven-badges.herokuapp.com`?

README.md
```diff
@@ -25,7 +25,10 @@ By making Event Hubs and Spark easier to use together, we hope this connector ma
...
+|Spark 2.1|azure-eventhubs-spark_2.11|[![Maven Central](https://img.shields.io/maven-central/v/com.microsoft.azure/azure-eventhubs-spark_2.11/2.2.0-PREVIEW.svg)](https://maven-badges.herokuapp.com/maven-central/com.microsoft.azure/azure-eventhubs-spark_2.11)|
+
+<sup>*Note:* Version 2.2.0 will supports Spark 2.1 as well.</sup>
```
jaceklaskowski: 1. s/will supports/will support
2. What's the purpose of the note? I don't understand it.

sabeegrewal: Well, 2.3.0 does *not* work with older versions of Spark while 2.2.0 will work with Spark 2.1 and Spark 2.2. The table should be sufficient though, so I'll drop the note for now.

README.md
sabeegrewal: Switching all to shields!

.github/CONTRIBUTING.md
sabeegrewal: Updated and links added 👍

sabeegrewal:
@jaceklaskowski, everything is looking cleaner - thanks again for the help! We're going to merge this now, however any additional feedback/suggestions are very welcomed! Hope to hear from you soon 👍

PR #274, opened 2018-03-14
Translate method optimization

### LINKED ISSUE #271: Connector fails to start when empty partitions exist
Currently, if your EventHub has empty partitions, you will get an error along the lines of:

`translate: result size X does not equal partition count Y`

To be fixed in 2.3.1.

### LINKED ISSUE #272: Re-write initialization code to avoid extra threads
If there are `N` partitions, we create `N` TCP connections, `N` AMQP Links, and `2 * N` threads to initialize a stream. This can be refactored to `1` TCP connection with `N` AMQP Links with `N` threads.

sabeegrewal:
Close #272 
Close #271 

This PR ensures `translate` only uses a single client, works on empty partitions (and static partitions), and is much faster due to reduced timeouts and reduced amqp link creations.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
@@ -242,11 +242,8 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
             receiver.setPrefetchCount(PrefetchCountMinimum)
             receiver.setReceiveTimeout(Duration.ofSeconds(5))
             val events = receiver.receiveSync(1) // get the first event that was received.
-            val event = events.iterator.next
-            result.put(partitionId, event.getSystemProperties.getSequenceNumber)
-          } catch {
-            case e: NullPointerException =>
-              logWarning("translate: failed to receive event.", e)
+            if (events == null || !events.iterator.hasNext) {
+              logWarning("translate: failed to receive event.")
               // No events to receive can happen in 2 cases:
               //   1. Receive from EndOfStream and no new events arrive
               //   2. Receive from an empty partition
@@ -256,10 +253,14 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
               } else {
                 result.put(partitionId, latest)
               }
-
-            case x: IllegalEntityException =>
-              logError("translate: IllegalEntityException. Consumer group may not exist.", x)
-              throw x
+            } else {
+              val event = events.iterator.next
+              result.put(partitionId, event.getSystemProperties.getSequenceNumber)
+            }
+          } catch {
+            case e: IllegalEntityException =>
+              logError("translate: IllegalEntityException. Consumer group may not exist.", e)
+              throw e
           } finally {
             if (receiver != null) {
               receiver.closeSync()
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
```diff
@@ -43,7 +44,13 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
 
   private implicit val formats = Serialization.formats(NoTypeHints)
 
-  private var client = ClientConnectionPool.borrowClient(ehConf)
+  private var _client: EventHubClient = _
+  private def client = {
+    if (_client == null) {
```
SreeramGarlapati: >if (_client == null) { [](start = 4, length = 22)

this is being accessed from multiple threads - this block will need synchronization. otherwise - this could lead to borrowing multiple clients.. but keeping track of only 1 of them.

core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
```diff
@@ -222,45 +225,38 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
...
-            }
-            newReceiver.closeSync()
-          } else {
+                                  positions.getOrElse(nAndP, defaultPos).convert)
+            receiver.setPrefetchCount(PrefetchCountMinimum)
+            receiver.setReceiveTimeout(Duration.ofSeconds(5))
+            val events = receiver.receiveSync(1) // get the first event that was received.
             val event = events.iterator.next
             result.put(partitionId, event.getSystemProperties.getSequenceNumber)
+            receiver.closeSync()
```
SreeramGarlapati: >receiver.closeSync() [](start = 12, length = 20)

move this to finally equivalent in scala.. 

Concern is: if receiveSync throws exception - receiver is not cleaned up.

core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
```diff
@@ -222,45 +225,38 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
...
-                newReceiver.getRuntimeInformation.getLastEnqueuedSequenceNumber
-              result.put(partitionId, lastSequenceNumber + 1)
-            }
-            newReceiver.closeSync()
-          } else {
+                                  positions.getOrElse(nAndP, defaultPos).convert)
+            receiver.setPrefetchCount(PrefetchCountMinimum)
+            receiver.setReceiveTimeout(Duration.ofSeconds(5))
+            val events = receiver.receiveSync(1) // get the first event that was received.
             val event = events.iterator.next
```
SreeramGarlapati: >val event = events.iterator.next [](start = 12, length = 32)

build safe gaurd for; if receiveSync returns null events.iterator would throw nullpointerex.

sabeegrewal: The intention here is to generate a NPE which should only happen from a receiver timeout. 

I was considering breaking up the catch block like so:

```scala
catch {
  case a: NullPointerException => 
    // handle null event case
  case b: IllegalEntityException =>
    // stop job
```
Thoughts on that approach?

SreeramGarlapati: I would recommended using `if else` as the control-flow logic than `try catch` ! :)
as we can simply check events == null - letting it throw `nullpointerex` is a bit counter intuitive :)

sabeegrewal: Ok, cool. So, it'll end up like this:

```scala
try {
  // try to receive an event, if it's null handle that accordingly
} catch {
  // explicitly catch IllegalEntityException, educate customer on consumer groups, kill the job
} finally { 
  // clean up 
}
```

Look good?

PR #279, opened 2018-03-16
Add `properties` to schema

### LINKED ISSUE #251: EventDataSchema: add EventData.properties - the application properties bag to the schema

sabeegrewal:
Add application properties from EventData in the exposed schema.

Close #251

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
+++ b/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
@@ -157,7 +157,7 @@ private[sql] object EventHubsSourceProvider extends Serializable {
         StructField("publisher", StringType),
         StructField("partitionKey", StringType),
         StructField("properties",
-                    MapType(StringType, StringType, valueContainsNull = true),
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
```diff
@@ -155,7 +155,10 @@ private[sql] object EventHubsSourceProvider extends Serializable {
         StructField("sequenceNumber", LongType),
         StructField("enqueuedTime", TimestampType),
         StructField("publisher", StringType),
-        StructField("partitionKey", StringType)
+        StructField("partitionKey", StringType),
+        StructField("properties",
+                    MapType(StringType, StringType, valueContainsNull = true),
```
SreeramGarlapati: is there a reason that you had to change from what EventData has: `Map<String, Object> getProperties();` !!?

One advantage with going with `Map<String,Object>` is users can send any type for `value` and transform it per their need.

sabeegrewal: `MapType` requires a `DataType`. `AnyDataType` is a `AbstractDataType` - seems `StringType` was the best fit.

Issue #259, opened 2018-03-07
Utilize Preferred Location to reuse pre-fetch queue across batches


sabeegrewal:
In Spark Streaming and Structured Streaming, partitions are consistently assigned to the same Executor. With this feature, we should make sure we maintain pre-fetch queues across batches (instead of creating and destroying them per batch).

PR #280, opened 2018-03-20
Parallel reader control

slyons:
Enables better core usage when cores > # of EH partitions


slyons:
@sabeegrewal Requesting review

sabeegrewal:
cc @sjkwak 

Hey Scott, we can have a meeting dedicated to this issue if you'd like - I'm fairly sure this isn't the direction we want to head in. 

When we create a receiver link, events start getting sent over immediately up to whatever the `PrefetchCount` is set to (even if you don't call `receive`). That prefetch buffer is maintained in-memory for the life of the receiver link (and it's thrown away once the receiver is closed). So, if we have 5 separate links, we would have 5 separate buffers. 

Now that a given partition is consistently assigned to the same machine, we want to start maintaining that in-memory buffer (i.e. that receiver link) beyond a single task (rather than creating more links per task). That way the buffer is always full and a `receive` call becomes O(1) since it's just a simple in-memory dequeue. The only tricky part is handling when executors are added/removed - we just have to make sure we don't skip or double-send any messages.

sabeegrewal:
Closing this for now. Very closely tied to #259. 

We need to have a long-lasting receiver on the executor so that all events are in-memory at the start of the task. Additionally, we need to build a "smart buffer" on top of this prefetch queue (something kind of like a `Map[Sequence Number, EventData]`). At that point, we can parallelize consumption from this buffer if we'd like to improve executor core utilization.

PR #281, opened 2018-03-20
fix synchronization bug in coonnection pool

tilumi:
Fix synchronization problem to prevent client may be created multiple times.


sabeegrewal:
Thanks!

PR #286, opened 2018-03-22
Add 'properties' to schema

### LINKED ISSUE #251: EventDataSchema: add EventData.properties - the application properties bag to the schema

sabeegrewal:
Add application properties from EventData in the exposed schema.

Close #251


sabeegrewal:
Ok, finally decided to close this for now. The simplest thing would be to use `ObjectType` in Spark. I'm going to look into how viable that is. The other option is User Defined Types in Structured Streaming which I know is planned. Both would be cleaner than this. Worst case, we have this ready in case somebody requests this feature before either of those make it to Spark.

sabeegrewal:
just rebasing 👍

PR #306, opened 2018-04-03
Change keys in singletons to be globally valid

### LINKED ISSUE #305: Singletons should lookup with namespace fqdn, name, and partition.
To avoid conflicts with other queries, singletons should have a `Map[ConnectionString, Resource]` to avoid conflicts.

sabeegrewal:
Change keys in singletons to use globally valid keys. Close #305

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -98,7 +98,11 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
 
   type MutableMap[A, B] = scala.collection.mutable.HashMap[A, B]
 
-  private[this] val receivers = new MutableMap[NameAndPartition, CachedEventHubsReceiver]()
+  private[this] val receivers = new MutableMap[String, CachedEventHubsReceiver]()
+
+  private def key(ehConf: EventHubsConf, nAndP: NameAndPartition): String = {
+    ehConf.connectionString + nAndP.partitionId
+  }
 
   private[eventhubs] override def receive(ehConf: EventHubsConf,
                                           nAndP: NameAndPartition,
@@ -106,7 +110,7 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
                                           batchSize: Int): EventData = {
     var receiver: CachedEventHubsReceiver = null
     receivers.synchronized {
-      receiver = receivers.getOrElseUpdate(nAndP, {
+      receiver = receivers.getOrElseUpdate(key(ehConf, nAndP), {
         CachedEventHubsReceiver(ehConf, nAndP, requestSeqNo)
       })
     }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -98,15 +98,19 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
 
   type MutableMap[A, B] = scala.collection.mutable.HashMap[A, B]
 
-  private[this] val receivers = new MutableMap[NameAndPartition, CachedEventHubsReceiver]()
+  private[this] val receivers = new MutableMap[String, CachedEventHubsReceiver]()
+
+  private def key(ehConf: EventHubsConf, nAndP: NameAndPartition): String = {
+    ehConf.connectionString + nAndP.partitionId
```
SreeramGarlapati: a) `ehConf.connectionString + ehConf.getOperationTimeout() + ehConf.getConsumerGroup() + nAndP.partitionId`
b) ensure applying the `ReceiveTimeout` corresponding to this `EHConf` after fetching the receiver ?

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -69,41 +69,42 @@ object ClientConnectionPool extends Logging {
...
+      val message = notInitializedMessage(key)
       throw new IllegalStateException(message)
     }
   }
 
-  private def get(name: String): ClientConnectionPool = pools.synchronized {
-    pools.getOrElse(name, {
-      val message = notInitializedMessage(name)
+  private def key(ehConf: EventHubsConf): String = {
+    ehConf.connectionString
```
SreeramGarlapati: consider normalizing the `connectionString`
- to handle the case where customers could give any `case` on `namespace` and `eventhub names` and pack them in any order into the connection string.

Probably, having this normalized while initializing `EHConf` is the way-to-go - to eliminate the need to validate this on every call.

sabeegrewal: I agree, I will add some normalization to both keys!

sabeegrewal:
@SreeramGarlapati, the connection string in `EventHubsConf` will always contain an `EntityPath`, so I don't see how adding timeouts will help keep the keys unique. I agree consumer group should be added for the cached receiver. 

Take a look...let me know if the keys make sense to you.

SreeramGarlapati:
The current change - Looks good.

Regarding timeout as key - I was trying to point out a corner case :D 
- lets say you have an `EHConf` - `ehConf1` with {ehName,connstr,cg1,opTimeout=30sec} and `ehConf2` with {ehName,connStr,cg2,opTimeout=45sec} - EventHubClient objects with different timeouts shouldn't be mixed up.

sabeegrewal:
I'm going to go ahead and merge - I think we should regard two `EventHubsConf`s with the same namespace and Event Hub name within  single query as an unsupported scenario for now.

PR #320, opened 2018-05-18
Cached receivers use epoch receivers

### LINKED ISSUE #317: Too many receivers are created if we 5+ cache misses in receiver cache
We'll solve this by switching to epoch receivers! That way only one receiver per partition will ever be cached.

sabeegrewal:
Close #317

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -27,6 +27,8 @@ import org.apache.spark.{ SparkEnv, TaskContext }
 import org.apache.spark.eventhubs.{ EventHubsConf, NameAndPartition, SequenceNumber }
 import org.apache.spark.internal.Logging
 
+import scala.util.{ Failure, Success, Try }
+
 private[spark] trait CachedReceiver {
   private[eventhubs] def receive(ehConf: EventHubsConf,
                                  nAndP: NameAndPartition,
@@ -65,11 +67,9 @@ private[client] class CachedEventHubsReceiver private (ehConf: EventHubsConf,
   }
 
   private def closeReceiver(): Unit = {
-    try {
-      _receiver.closeSync()
-      _receiver = null
-    } catch {
-      case e: Exception => logInfo("closeSync failed on cached receiver.", e)
+    Try(_receiver.closeSync()) match {
+      case Success(_) => _receiver = null
+      case Failure(e) => logInfo("closeSync failed in cached receiver.", e)
     }
   }
 
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -17,12 +17,7 @@
 
 package org.apache.spark.eventhubs.client
 
-import com.microsoft.azure.eventhubs.{
-  EventData,
-  EventHubClient,
-  PartitionReceiver,
-  ReceiverOptions
-}
+import com.microsoft.azure.eventhubs._
 import org.apache.spark.{ SparkEnv, TaskContext }
 import org.apache.spark.eventhubs.{ EventHubsConf, NameAndPartition, SequenceNumber }
 import org.apache.spark.internal.Logging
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -83,6 +85,7 @@ private[client] class CachedEventHubsReceiver private (ehConf: EventHubsConf,
       logWarning(
         s"$requestSeqNo did not match ${event.getSystemProperties.getSequenceNumber}." +
           s"Recreating receiver for $nAndP")
+      _receiver.closeSync()
```
SreeramGarlapati: >_receiver.closeSync() [](start = 6, length = 21)

if close fails - we still want to go ahead and create the new Receiver..
So, lets absorb any non-fatal errors - and log'em.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -55,10 +55,12 @@ private[client] class CachedEventHubsReceiver private (ehConf: EventHubsConf,
     receiverOptions.setReceiverRuntimeMetricEnabled(false)
     receiverOptions.setIdentifier(
       s"spark-${SparkEnv.get.executorId}-${TaskContext.get.taskAttemptId}")
-    _receiver = client.createReceiverSync(consumerGroup,
-                                          nAndP.partitionId.toString,
-                                          EventPosition.fromSequenceNumber(requestSeqNo).convert,
-                                          receiverOptions)
+    _receiver = client.createEpochReceiverSync(
```
SreeramGarlapati: >createEpochReceiverSync [](start = 23, length = 23)

By changing from standard receiver to epoch receiver - impact for customers is that:
In earlier model they could be running multiple Jobs - by using $default consumerGroup
With Epoch receivers, those jobs would start seeing failures.
eval, if we can take proactive measure here:
1. announce this CHANGE of behv. loud and clear in release notes.
2. add faq - for ReceiverDisconnectedException with epoch errors.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -83,6 +94,7 @@ private[client] class CachedEventHubsReceiver private (ehConf: EventHubsConf,
       logWarning(
         s"$requestSeqNo did not match ${event.getSystemProperties.getSequenceNumber}." +
           s"Recreating receiver for $nAndP")
+      closeReceiver()
```
SreeramGarlapati: if close fails - we still want to go ahead and create the new Receiver..
So, lets absorb non-fatal errors - and log'em.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
SreeramGarlapati: One of the ways to solve this could be - to add this possibility of 1 spark job trumping another - in via traces.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
sabeegrewal: what change? take a look at the closeReceiver method I added - i log the error there.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
sabeegrewal: this is actually a pretty big issue, we should talk about this more before merging

sabeegrewal:
Ok, before merging:

- [ ] catch epoch disconnection errors on receive calls and pass clearer messaging
- [ ] remove and replace try catches
- [ ] add an FAQ 
- [ ] rename pyspark folder

core/src/main/scala/org/apache/spark/eventhubs/package.scala
```diff
@@ -39,6 +39,7 @@ package object eventhubs {
   val DefaultFailOnDataLoss = "true"
   val DefaultUseSimulatedClient = "false"
   val StartingSequenceNumber = 0L
+  val DefaultEpoch = 0L
```
SreeramGarlapati: tip: use a small & unique number like 1761 
 - from the error messages you can tell - another spark job trumped the receiver

PR #328, opened 2018-05-31
add properties to schema

### LINKED ISSUE #250: EventData schema: Expose SystemProperties/Properties
Exposing `SystemProperties` will help us deliver amqp message annotations to EventDataSchema. `IOT-client` relies on this feature to deliver `device-id` to their customers.

```
def eventHubsSchema: StructType = {
    StructType(
      Seq(
        StructField("body", BinaryType),
        StructField("systemProperties", MapType(StringType, AnyDataType())),
     ...... 
      ))
  }
```

As I tried to highlight - 
1) we should expose this as a key value pair of string, object and 
2) remove the 3 standard sysproperties (offset, ...) to de-duplicate the payload size

sabeegrewal:
close #250 

This adds properties to the Spark SQL schema in the form of a `Map[String, String]`. Where the values are all serialized as JSON strings. This uses json4s and jackson (as does Spark).

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
+++ b/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
@@ -156,7 +156,8 @@ private[sql] object EventHubsSourceProvider extends Serializable {
         StructField("sequenceNumber", LongType),
         StructField("enqueuedTime", TimestampType),
         StructField("publisher", StringType),
-        StructField("partitionKey", StringType)
+        StructField("partitionKey", StringType),
+        StructField("properties", MapType(StringType, StringType), nullable = true)
       ))
   }
 }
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
```diff
@@ -156,7 +156,8 @@ private[sql] object EventHubsSourceProvider extends Serializable {
         StructField("sequenceNumber", LongType),
         StructField("enqueuedTime", TimestampType),
         StructField("publisher", StringType),
-        StructField("partitionKey", StringType)
+        StructField("partitionKey", StringType),
+        StructField("properties", MapType(StringType, StringType), nullable = true)
```
SreeramGarlapati: Is it possible to provide the below different types of Schemas
1) Schema without properties
2) schema with properties - which serializes property values as strings
3) schema with properties - which does it this way - json way
and then, be able to choose schema - based on the Configuration. Perhaps, next step in this direction is to keep the schema part extensible in the spark adapter.

My concern here is that:
- all users - irrespective of whether they have just a single type of property value - need to deal with deserializing json; it will be good - if we have a way to support simple types easily and then provide ways to achieve these advanced serialization scenarios.
- Serialization errors, if exists, will pop-up - for the users who doesn't really care about properties for their scenario.

sabeegrewal:
I'm going to add some better tests to make sure all amqp properties are serialized correctly 👍

PR #329, opened 2018-06-01
Trim EventHubsConf before sending to Spark Executors

### LINKED ISSUE #258: Only pass needed parameters to executors
Instead of passing the entire `EventHubsConf`, we should only pass the features that we use on the executor.

sabeegrewal:
Removes all the options that are never used by Spark Executors. Just a simple optimization so we aren't sending too much info across the Spark cluster. 

close #258

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -71,6 +71,11 @@ final class EventHubsConf private (private val connectionStr: String)
     this
   }
 
+  private def remove(key: String): EventHubsConf = {
+    settings.remove(key)
+    this
+  }
+
   private[spark] def apply(key: String): String = {
     get(key).get
   }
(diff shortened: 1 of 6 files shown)


SreeramGarlapati:
Hi @sabeegrewal - I was looking into this and saw this got merged. so, giving my thoughts here.
intent of this change was good. however, I believe - approach could be different. consider the following:

Consider having the `inclusion list` instead of the `exclusion list`: with the current approach, everytime a new config is added to `EventHubConf` - it would be flowing thru - unless it is added to this list - which a new developer will have to be reminded.. with this change - new configs will be added - only if they are needed at the `Executors`, as, developers are gonna see failures if not!

In this change - if I understand `scala` correctly - trim method removes the entries from the actual `property bag' in `EventHubConf` - consider an approach which retains the original config as well. Being able to re-read values - is essential.

sabeegrewal:
Hey @SreeramGarlapati - you're definitely right. Made the change in #335.

PR #335, opened 2018-06-06
Trim executor options based on included options

sabeegrewal:
Trimming options based on included options (not excluded options). When new params are added, they'll need to be explicitly added to the `include` list in order to be passed to executors. This ensures we don't accidentally pass unneeded options to executors by mistake. 👍

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -389,14 +389,15 @@ object EventHubsConf extends Logging {
    * @return trimmed [[EventHubsConf]]
    */
   private[spark] def trim(ehConf: EventHubsConf): EventHubsConf = {
+    // These are the options needed by Spark executors
+    val include = Seq("eventhubs.connectionString",
+                      "eventhubs.consumerGroup",
+                      "eventhubs.receiverTimeout",
+                      "eventhubs.operationTimeout",
+                      "useSimulatedClient")
+
+    ehConf.settings.asScala.filter(tuple => include.contains(tuple._1))
     ehConf
-      .remove("eventhubs.startingPosition")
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -389,14 +389,15 @@ object EventHubsConf extends Logging {
...
    */
   private[spark] def trim(ehConf: EventHubsConf): EventHubsConf = {
+    // These are the options needed by Spark executors
+    val include = Seq("eventhubs.connectionString",
+                      "eventhubs.consumerGroup",
+                      "eventhubs.receiverTimeout",
+                      "eventhubs.operationTimeout",
+                      "useSimulatedClient")
+
+    ehConf.settings.asScala.filter(tuple => include.contains(tuple._1))
```
SreeramGarlapati: u intend to return the filtered list right? shouldn't this be the last line of this method as opposed to `ehConf` ?

```
ehConf.settings.asScala.filter(tuple => include.contains(tuple._1))
```

sabeegrewal: No, we're returning the `EventHubsConf` which holds the internal list. So, we filter the list, and then return the ehConf which wraps it.

SreeramGarlapati: do you mean `filter` modifies the original collection - i doubt it!!
It should return a new settings map with the filtered collection - after running your lamda.

if the trim method is called infrequently & not called in hot path - something like - for every batch - having it to return a copy of settings should be fine too.

sabeegrewal: Ah shoot, `retain` modifies the original collection. I'll update it...

PR #341, opened 2018-06-14
Translate redo and parallelize getOffset

sabeegrewal:
- Translate has been rewritten. Now - in most cases - receivers are no longer created. If `EndOfStream` or `StartOfStream` are passed by the user, then we just cut to the chase and call `getLatest` or `getEarliest` to get the sequence number (rather than creating a receiver). 

- getOffset now does the `getRunTimeInfo` calls in parallel

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.eventhubs.client
 
-import java.util.concurrent.CompletableFuture
-
 import com.microsoft.azure.eventhubs._
 import com.microsoft.azure.eventhubs.impl.EventHubClientImpl
 import org.apache.spark.eventhubs.EventHubsConf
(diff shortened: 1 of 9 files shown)


sabeegrewal:
At some point we need to wait for the asynchronous requests in the Java client to complete. This PR is to 1) make all the requests in parallel and 2) make sure we don't make a thread wait until the we have to wait. 

Currently, we immediately call `get` which causes our thread to hold until the `CompletableFuture` completes. Instead, we'll use Scala `Futures` to keep service requests async until we have to wait for the service to respond. This shows significant speed up for high partition scenarios. 

More work/experimentation is going on, but I'm keeping the PR open in case anyone is interested.

core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
```diff
@@ -220,82 +278,61 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
...
-              } else {
-                result.put(partitionId, latest)
+    val futures = for ((nAndP, pos) <- needsTranslation)
+      yield
+        pos.offset match {
+          case StartOfStream => (nAndP.partitionId, earliestSeqNoF(nAndP.partitionId))
+          case EndOfStream   => (nAndP.partitionId, latestSeqNoF(nAndP.partitionId))
+          case _ =>
+            val receiver = toScala(
+              client.createEpochReceiver(consumerGroup,
```
sabeegrewal: add retry here

PR #359, opened 2018-07-18
Simplify configuration and creation of the client

Fokko:


Fokko:
Still WIP, need to verify if it works on Eventhub itself.

Did some minor refactoring:
- Made the `clientFactory` static so we don't need to pass this constructor around

Fokko:
@sabeegrewal Let me know what you think

sabeegrewal:
@Fokko still needs to look into this in more detail - I'll do a review tomorrow as well. Thanks for the PR! 

The worry with this change is that if you're stream is going so slowly that you're constantly missing events, then just dropping them out of your stream isn't a good approach. I was more on the side of failing these jobs because, to me, running into this issue indicates a poorly-tuned pipeline. What do you think about that? Curious to hear your thoughts. 

Also, what about the case where the whole range is dropped from Event Hubs? E.g. the `untilSeqNo` doesn't exist in the Event Hub either.

Fokko:
Let me think about it, not sure if this is the right approach. The `getPartitions` will also query the API every time, which is quite expensive.

sabeegrewal:
Right - let me know what you think. Happy to continue this conversation!

Fokko:
@sabeegrewal I've reverted the changes. The main problem was that querying the API is too expensive to do this every time `getPartitions` is called. I think correcting the minimal offsets in the factory method might be an option:
https://github.com/Azure/azure-event-hubs-spark/blob/master/core/src/main/scala/org/apache/spark/eventhubs/EventHubsUtils.scala#L72

But right now I don't have the problem anymore since the stream is up to date. I've created a new consumer group to replicate the error. I'll keep you posted.

In the meantime I think the refactoring that was involved, would still be valuable to merge to master. The main change is that the factory method, that initializes the client is now a static method: https://github.com/Azure/azure-event-hubs-spark/pull/359/files#diff-7e9b29ae3b93a01ae6736195c5d98795R147
Since we already have the config-map, we can determine based on that which client needs to be used.

sabeegrewal:
Hey @Fokko, I'm with you on your last post. A lot of my work recently has been reducing network i/o as much as possible which is why I'm always a little hesitant about this sort of stuff. With that said, I'm definitely happy to keep discussing it! 

For this PR, is it safe to remove the WIP labels and review it to be merged? Let me know and I'll spend some time with it. And, as always, thanks again for your help, it's much appreciated!

Fokko:
Thanks for the quick reply @sabeegrewal  This is ready to be reviewed.

I fully agree. Network I/O is expensive and should be kept to a minimum. I still have some ideas of optimizing this, but I need some time to come up with some actual benchmark. I've removed the additional API calls from the PR for now. The changes in this PR is reducing the complexity a bit.

sabeegrewal:
Sounds goood. Feel free to contact me on the gitter if you wanna chat more. I'll get the reviewed/merged soon 👍

Fokko:
Cool, thanks! Let me know if you have any further questions.

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -392,26 +392,6 @@ object EventHubsConf extends Logging {
     Serialization.write[T](value)
   }
 
-  /**
```
sabeegrewal: Are the tests still failing with this included? Is that the only motivation for removing this? 

The intention was the reduce the payload that is serialized and sent to executors (for example people with hundreds of partitions **and** they have `startingPositions` set).

pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <groupId>com.microsoft.azure</groupId>
   <artifactId>azure-eventhubs-spark-parent_2.11</artifactId>
-  <version>2.3.2</version>
+  <version>2.3.3</version>
```
sabeegrewal: `2.3.3-SNAPSHOT` for this one too. (It's really my bad for not updating it, so thanks!)

core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
```diff
@@ -62,14 +61,7 @@ private[sql] class EventHubsSourceProvider
     EventHubsClient.userAgent =
       s"Structured-Streaming-${sqlContext.sparkSession.sparkContext.version}"
 
-    val caseInsensitiveParameters = parameters.map {
```
sabeegrewal: What is the thought process for removing these? I'm pretty sure all the lookups assume lowercase keys.

Fokko: I would assume that people would supply the parameters in lowercase anyway. This logic might add a big of unpredictable behaviour. I would suggest to put it here in the factory, to keep everything on a single place:
https://github.com/Fokko/azure-event-hubs-spark/blob/fd-fix-event-hub-offset/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala#L147

What do you think?

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
Fokko: Let me check if I can come up with a test.

Fokko: I see, it gets serialized and send to the executor in here: https://github.com/Azure/azure-event-hubs-spark/blob/master/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsWriter.scala#L73

core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
Fokko: I've moved all the `.toLowerCase()` to the `EventHubsConf` to keep everything in one single place 👍

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
Fokko: I've restored the trimmed configs by using `.trimmed`

core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala
sabeegrewal: Ok, cool - that's definitely the right move.

PR #360, opened 2018-07-23
Split the classes from EventHubsTestUtils.scala in seperate files

Fokko:
The EventHubsTestUtils.scala was very long, therefore it is better to split the classes/objects into seperate files. I didn't change any of the content of the files.


core/src/main/scala/org/apache/spark/eventhubs/utils/SimulatedCachedReceiver.scala
```diff
@@ -0,0 +1,29 @@
+package org.apache.spark.eventhubs.utils
```
sabeegrewal: Could you add the apache header to all these new files?

sabeegrewal:
I agree this should be done - just a minor change + resolving the minor merge conflicts and I'll get this merged! Thanks for the contributions :)

core/src/main/scala/org/apache/spark/eventhubs/utils/SimulatedCachedReceiver.scala
Fokko: Good point, I've added the headers.

Fokko:
@sabeegrewal I've rebased onto master.

sabeegrewal:
@Fokko thanks man :) great stuff, I appreciate it!

PR #369, opened 2018-08-02
In translate, redo receive when it returns null

### LINKED ISSUE #357: A future can return null in EventHubsClient.translate
I'm seeing some unexpected behaviour since updating from 2.3.1 to 2.3.2 which I think is related to the cached receivers. 

I am running a pyspark job which fetches a dataframe from my EventHub, both `startingPosition` end `endingPosition` properties are set with a `enqueuedTime` some processing on the data is done and finally saved. 
First time I run the job all is fine, when I run the same job a second time however (after the first job is finished but on the the same cluster) I get following error  on:
`spark.read.format("eventhubs").options(**eh_conf).load()`
```
 java.lang.NullPointerException
	at org.apache.spark.eventhubs.client.EventHubsClient$$anonfun$9$$anonfun$12.apply(EventHubsClient.scala:246)
	at org.apache.spark.eventhubs.client.EventHubsClient$$anonfun$9$$anonfun$12.apply(EventHubsClient.scala:245)
	at scala.util.Success$$anonfun$map$1.apply(Try.scala:237)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Success.map(Try.scala:237)
	at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
	at scala.concurrent.Future$$anonfun$map$1.apply(Future.scala:237)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

This only is happening when using version 2.3.2, the same job with the 2.3.1 library runs fine on every subsequent run, always returning the same output.

I am using Spark version 2.3, I get the same behaviour when running on a Databricks interactive cluster and running pyspark in local mode.

sabeegrewal:
This fixes a bug in the `translate` method. Previously, when we called `receive`, we could return null which led to a `NullPointerException`. Now, the receive call guarantees we get an `EventData` before continuing. We keep this operation on the Scala Execution Context. 

close #357

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -352,7 +352,7 @@ final class EventHubsConf private (private val connectionStr: String)
   }
 
   /**
-   * Set the operation timeout. We will retry failures when contacting the
+   * Set the operation timeout. We will retryJava failures when contacting the
    * EventHubs service for the length of this timeout.
    * Default: [[DefaultOperationTimeout]]
    *
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -147,6 +147,7 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
                                           nAndP: NameAndPartition,
                                           requestSeqNo: SequenceNumber,
                                           batchSize: Int): EventData = {
+    logInfo(s"EventHubsCachedReceiver look up. For ${key(ehConf, nAndP)}")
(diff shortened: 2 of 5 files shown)


sabeegrewal:
Going to try to write this without using `blocking`

sabeegrewal:
K, done! `retryNotNull` will be used to build out async receives throughout the connector in an upcoming PR 👍

Issue #367, opened 2018-08-01
AmqpException during streaming 


sakthig:
I am consuming IoThub data using spark.readStream. I keep getting these below errors

```
2018-08-01 07:17:37 WARN  TaskSetManager:66 - Lost task 1.1 in stage 9.0 (TID 1425, 172.31.14.123, executor 1): com.microsoft.azure.eventhubs.EventHubException: com.microsoft.azure.eventhubs.impl.AmqpException: The service was unable to process the request; please retry the operation. For more information on exception types and proper exception handling, please refer to http://go.microsoft.com/fwlink/?LinkId=761101 Reference:892b4926-d06f-4cc5-8ba8-9eb2f77a5b20, TrackingId:4bbd1e9d0002017000100a4f5b615e91_G1_B3, SystemTracker:ihsuprodmares019dednamespace:eventhub:iothub-ehub-sensinerve-355077-6a40c0307a~32766|analytics, Timestamp:8/1/2018 7:17:37 AM, errorContext[NS: ihsuprodmares019dednamespace.servicebus.windows.net, PATH: iothub-ehub-sensinerve-355077-6a40c0307a/ConsumerGroups/analytics/Partitions/1, REFERENCE_ID: 1b5455_94d2_G1_1533107847267, PREFETCH_COUNT: 2000, LINK_CREDIT: 2000, PREFETCH_Q_LEN: 0]
	at com.microsoft.azure.eventhubs.impl.ExceptionUtil.toException(ExceptionUtil.java:39)
	at com.microsoft.azure.eventhubs.impl.MessageReceiver.onClose(MessageReceiver.java:620)
	at com.microsoft.azure.eventhubs.impl.BaseLinkHandler.processOnClose(BaseLinkHandler.java:50)
	at com.microsoft.azure.eventhubs.impl.BaseLinkHandler.handleRemoteLinkClosed(BaseLinkHandler.java:70)
	at com.microsoft.azure.eventhubs.impl.BaseLinkHandler.onLinkRemoteClose(BaseLinkHandler.java:36)
	at org.apache.qpid.proton.engine.BaseHandler.handle(BaseHandler.java:176)
	at org.apache.qpid.proton.engine.impl.EventImpl.dispatch(EventImpl.java:108)
	at org.apache.qpid.proton.reactor.impl.ReactorImpl.dispatch(ReactorImpl.java:324)
[...]

Issue #363, opened 2018-07-24
ReactorDispatcher instance is closed | MessageReceiver instance is closed.


vijayupadya:
We are frequently hitting the below error when reading from the eventhub. We see this on both DataBricks and HDInsights.

Version: azure-eventhubs-spark_2.11-2.3.2

**Callstack:**

java.util.concurrent.RejectedExecutionException: ReactorDispatcher instance is closed.
	at com.microsoft.azure.eventhubs.impl.ReactorDispatcher.throwIfSchedulerError(ReactorDispatcher.java:81)
	at com.microsoft.azure.eventhubs.impl.ReactorDispatcher.invoke(ReactorDispatcher.java:57)
	at com.microsoft.azure.eventhubs.impl.MessagingFactory.scheduleOnReactorThread(MessagingFactory.java:375)
	at com.microsoft.azure.eventhubs.impl.MessageReceiver.receive(MessageReceiver.java:268)
	at com.microsoft.azure.eventhubs.impl.PartitionReceiverImpl.receive(PartitionReceiverImpl.java:142)
	at com.microsoft.azure.eventhubs.PartitionReceiver.lambda$receiveSync$0(PartitionReceiver.java:89)
	at com.microsoft.azure.eventhubs.impl.ExceptionUtil.sync(ExceptionUtil.java:179)
	at com.microsoft.azure.eventhubs.PartitionReceiver.receiveSync(PartitionReceiver.java:89)
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver.liftedTree1$1(CachedEventHubsReceiver.scala:103)
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver.org$apache$spark$eventhubs$client$CachedEventHubsReceiver$$receive(CachedEventHubsReceiver.scala:102)
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver$.receive(CachedEventHubsReceiver.scala:157)
	at org.apache.spark.eventhubs.rdd.EventHubsRDD$EventHubsRDDIterator.next(EventHubsRDD.scala:147)
	at org.apache.spark.eventhubs.rdd.EventHubsRDD$EventHubsRDDIterator.next(EventHubsRDD.scala:128)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:409)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
[...]

PR #377, opened 2018-08-08
Cached receivers are async

sabeegrewal:
This re-works cached receivers. All receivers and receive calls are done asynchronously. All receive calls are retried as well and they're retried until a non-null event is returned. 

When we hand off the events to Spark, we make sure the data is in order by sequence number. 

#363 #367

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -18,17 +18,21 @@
 package org.apache.spark.eventhubs.client
 
 import com.microsoft.azure.eventhubs._
+import org.apache.spark.eventhubs.utils.RetryUtils.{ retryJava, retryNotNull }
 import org.apache.spark.{ SparkEnv, TaskContext }
 import org.apache.spark.eventhubs.{ EventHubsConf, NameAndPartition, SequenceNumber }
 import org.apache.spark.internal.Logging
 
-import scala.util.{ Failure, Success, Try }
+import scala.concurrent.{ Await, Future }
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.collection.JavaConverters._
+import scala.util.Success
 
 private[spark] trait CachedReceiver {
   private[eventhubs] def receive(ehConf: EventHubsConf,
                                  nAndP: NameAndPartition,
                                  requestSeqNo: SequenceNumber,
-                                 batchSize: Int): EventData
+                                 batchSize: Int): Iterator[EventData]
 }
 
 /**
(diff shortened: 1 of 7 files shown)


sabeegrewal:
Ok, LGTM - reviews are welcome :)

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -48,83 +52,94 @@ private[spark] trait CachedReceiver {
...
+                                                       startSeqNo: SequenceNumber)
     extends Logging {
 
   import org.apache.spark.eventhubs._
 
   private lazy val client: EventHubClient = ClientConnectionPool.borrowClient(ehConf)
 
+  private var _receiver: Future[PartitionReceiver] = _
+  private def receiver: Future[PartitionReceiver] = {
+    if (_receiver == null) {
```
Fokko: When could the receiver be null? It is created right at the initialization of the class. I'm not super familiar with the execution path of Spark, but does it re-use closed receivers? This would be an anti pattern.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -48,83 +52,94 @@ private[spark] trait CachedReceiver {
...
-      assert(requestSeqNo == event.getSystemProperties.getSequenceNumber,
-             errWrongSeqNo(requestSeqNo, event.getSystemProperties.getSequenceNumber))
+      receiveOne
+    } else {
+      Future { event }
+    }
+  }
+
+  private def receive(requestSeqNo: SequenceNumber, batchSize: Int): Iterator[EventData] = {
+    val first = checkCursor(requestSeqNo)
```
Fokko: Not sure if we need all this future logic here. Also doubting if it would make it more performant.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -160,8 +175,6 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
   def apply(ehConf: EventHubsConf,
             nAndP: NameAndPartition,
             startSeqNo: SequenceNumber): CachedEventHubsReceiver = {
-    val cr = new CachedEventHubsReceiver(ehConf, nAndP)
-    cr.createReceiver(startSeqNo)
-    cr
+    new CachedEventHubsReceiver(ehConf, nAndP, startSeqNo)
```
Fokko: Nice one

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
Fokko: Maybe we can do something smart and create a wrapper around it?

sabeegrewal: **The problem:** Let's say we have our receiver for partition `x` cached on executor `y`. We consume up to sequence number 2000. Then an executor is added, so now tasks for partition `x` are assigned to executor `z`. There's a cache miss and a new cache is made. Let's say we consume until sequence number 3000. Then, just after that, the newly added executor is removed by Spark. Now, partition `x` tasks are back to executor `y`. There's a cache hit. We expect to consume sequence number 3001, but the next queued event is 2001 - that's where we left off in this cache. 

My current solution is to close the receiver and create a new one at the right spot. Though, after thinking about this, a better solution may be to just consume events until we get to the expected sequence number.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
sabeegrewal: See "the problem" above. There's a slight chance that the cursor can be in the wrong spot!

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
sabeegrewal: Yea, since the seqNos (and receive calls) aren't being made one at a time any more, there's actually more cleaning I can do. I'll try to get rid of the `null` receiver and clean this a bit more tomorrow.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -48,83 +52,88 @@ private[spark] trait CachedReceiver {
...
-        logInfo("closeSync failed in cached receiver.", e)
-        _receiver = null
-    }
-  }
-
-  private var _receiver: PartitionReceiver = _
-  private def receiver: PartitionReceiver = {
-    if (_receiver == null) {
-      throw new IllegalStateException(s"No receiver for $nAndP")
+    val r = retryJava(
```
Fokko: Small nit, but I always try to avoid 1 letter variables.

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -48,83 +52,88 @@ private[spark] trait CachedReceiver {
...
-        }
+  private def receiveOne: Future[Iterable[EventData]] = {
+    receiver
+      .flatMap { r =>
+        retryNotNull(r.receive(1), "CachedReceiver receive call")
       }
-      event = i.iterator.next
-      event
-    }
+      .map { e =>
```
Fokko: I would simplify this as `.map(_.asScala)`

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -48,83 +52,88 @@ private[spark] trait CachedReceiver {
...
+    val event = Await.result(receiveOne, InternalOperationTimeout)
+    val receivedSeqNo = event.head.getSystemProperties.getSequenceNumber
+    if (receivedSeqNo < requestSeqNo) {
+      // If the receivedSeqNo is less than the requestSeqNo,
+      // then some batches must have been completed on a
+      // separate executor. This can happen when an executor
+      // is added then removed.
       logWarning(
-        s"$requestSeqNo did not match ${event.getSystemProperties.getSequenceNumber}." +
+        s"$requestSeqNo did not match $receivedSeqNo." +
```
Fokko: Move this comment onto one line.

sabeegrewal:
Thanks @Fokko :) I think I addressed everything! Should be merged in today.

PR #385, opened 2018-09-10
Add the consumer group to the exception

Fokko:
When this happens, we want to know which consumer group the Apache Spark job is using to make debugging easier.


Fokko:
Thanks @sabeegrewal

PR #394, opened 2018-10-03
Receive calls one at a time to avoid overloading thread pool

sabeegrewal:
In this PR:
* `receiveOne` is called one at a time. Specifically, we only issue one async receive call at a time, and we don't issue the *next* receive call until the *previous* is completed. This should remove load from the underlying client thread pool for large batches, and, therefore, ensure fairer use of resources across the cluster.
* `receiveOne` now accepts a specialized message to improve tracing. All receive calls have a specific message so we can identify failures more easily.


sabeegrewal:
bumped the version 👍 will merge once tests run

PR #395, opened 2018-10-06
add all system properties to schema

### LINKED ISSUE #366: Make IoT Hub system properties available
IoT Hub system properties (https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-messages-construct), e.g. ConnectionDeviceID from messages sent via iot-hub, are not available in properties. Should be available in column properties.

Library: com.microsoft.azure:azure-eventhubs-spark_2.11:2.3.2
Spark 2.3.1, Scala 2.11 (Azure Databricks Runtime 4.2)

sabeegrewal:
close #366 

In this PR we add all System Properties to the schema. This includes all IotHub related properties.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/package.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/package.scala
@@ -50,6 +50,10 @@ package object eventhubs {
   val RetryCount = 3
   val InternalOperationTimeout: FiniteDuration = 300.seconds
 
+  val OffsetAnnotation = "x-opt-offset"
+  val EnqueuedTimeAnnotation = "x-opt-enqueued-time"
+  val SequenceNumberAnnotation = "x-opt-sequence-number"
+
   type PartitionId = Int
   val PartitionId: Int.type = Int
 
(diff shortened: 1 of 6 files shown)


sabeegrewal:
@SreeramGarlapati @sjkwak 
In this PR, `sequence number`, `offset`, and `enqueue time` are now duplicated. Do you think that's okay? I have a view options that would handle the duplicates as well (though it doesn't strike me as a big deal). I'll catch you guys on Monday :)

SreeramGarlapati:
@sabeegrewal - consider not offering them in the systemproperties bag - as you already have these 3 properties available directly on the schema. API design wise - its always recommended to have 1 way to get stuff done.

sabeegrewal:
I agree, but then we have to unpack the system properties bag 1 key at a time. I don't think that's a good option. 

What about removing `sequence number`, `offset`, and `enqueue time` from the system properties map? That way there's only one way to get them AND we don't have to concern ourselves with all the potential keys in the system properties bag. @SreeramGarlapati

PR #406, opened 2018-11-05
Update logging for EventHubsConf class

### LINKED ISSUE #403: Remove the log of configuration value when it is overwritten
- What issue are you trying to solve?
Spark job log contains connection string including the access key, for example:

18/10/16 04:26:50 WARN EventHubsConf: eventhubs.connectionstring has already been set to Endpoint=sb://***.servicebus.windows.net;EntityPath=t***;SharedAccessKeyName=***;SharedAccessKey=***. Overwriting with Endpoint=sb://***.servicebus.windows.net;EntityPath=***;SharedAccessKeyName=***;SharedAccessKey=***

we want to remove the raw value of connection string in this log in case of leaking the secret keys.

- How do you want to solve it? 
The log comes from 
https://github.com/Azure/azure-event-hubs-spark/blob/master/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
around line 68:
    if (self.get(lowerKey).isDefined) {
      logWarning(s"$key has already been set to ${self.get(key).get}. Overwriting with $value")
    }

we may update it to logWarning(s"$key has already been set.")

Also it would be better to see if the value is the same first, and skip overwriting if the new value is the same as the old.

sjkwak:
close #403

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -64,7 +64,7 @@ final class EventHubsConf private (private val connectionStr: String)
 
     val lowerKey = key.toLowerCase
 
-    if (self.get(lowerKey).isDefined) {
+    if (self.get(lowerKey).isDefined && lowerKey != ConnectionStringKey.toLowerCase) {
       logWarning(s"$key has already been set to ${self.get(key).get}. Overwriting with $value")
     }
 


core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -64,7 +64,7 @@ final class EventHubsConf private (private val connectionStr: String)
 
     val lowerKey = key.toLowerCase
 
-    if (self.get(lowerKey).isDefined) {
+    if (self.get(lowerKey).isDefined && lowerKey != ConnectionStringKey.toLowerCase) {
       logWarning(s"$key has already been set to ${self.get(key).get}. Overwriting with $value")
```
sabeegrewal: don't print `self.get(key).get` - that needs to be removed!

sabeegrewal:
close #403

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
sjkwak: Sure. Let me send fix that as well.

sabeegrewal: you can remove `&& lowerKey != ConnectionStringKey.toLowerCase` as well (if you stop logging the value).

sjkwak: For a fix, I think we need to log existing and new value for any update as long as config key is not a connection string. For connection string, we will stop logging the connection string. Do you agree?

sabeegrewal: That's fine - if that's what is desired, then you can leave it as is. I thought it'd be better to just never log old/new values and print out a consistent message for all cases.

PR #408, opened 2018-11-10
[WIP] Issues with seqNo and Eventhub Retention

Fokko:
It feels like we're having some issues with the retention of the Event Hub.
When starting a Spark job that reads from a consumer group that already has some pruned messages
because then run out of retention.

```
java.lang.IllegalStateException: In partition 1 of eventhubdev, with consumer group consumergroup-1,
request seqNo 1921738 is less than the received seqNo 1924294.
The earliest seqNo is 1924293 and the last seqNo is 1944993
```

First I need to fix another issue with the Simulator, which I'll do in another PR to keep everything nicely separated. Currently, the latest seqNo is implemented by taking the `size()` of the messages within a certain partition. This is not right, we want to take the max of the seqNo of the actual messages:
https://github.com/Azure/azure-event-hubs-spark/blob/61aba38334f0bf13a7c2bbee0fafd0b7653e687e/core/src/main/scala/org/apache/spark/eventhubs/utils/SimulatedEventHubs.scala#L252

Thanks for contributing! We appreciate it :) 

For a Pull Request to be accepted, you must:
- Run scalafmt on your code using the `.scalafmt.conf` present in this project
- All tests must pass when you run `mvn clean test` 

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/utils/EventHubsTestUtils.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/utils/EventHubsTestUtils.scala
@@ -20,19 +20,13 @@ package org.apache.spark.eventhubs.utils
 import java.util.Date
 
 import com.microsoft.azure.eventhubs.EventData
-import com.microsoft.azure.eventhubs.impl.AmqpConstants.{
-  ENQUEUED_TIME_UTC,
-  OFFSET,
-  SEQUENCE_NUMBER
-}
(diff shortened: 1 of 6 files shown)


Fokko:
Please merge #409 first

sabeegrewal:
@Fokko sorry for the hold up...can you rebase when you get the chance? Also, is this ready for review? 

as always, thanks for taking the time.

sabeegrewal:
Hey, I'm going to close this for now. If you continue to have issues after the upcoming release, then we should revisit this PR.

PR #409, opened 2018-11-10
Take the actual sequence numbers, instead of relying on the underlying Seq

Fokko:
For the tests we want to use the actual sequence number instead of using the size of the underlying Seq. In real world, the consumer group of the event hub might nog start at zero, but at a high number.


Fokko:
Required by #408

PR #415, opened 2018-11-28
Use OperationTimeout config value instead of fixed timeout when specifying timeout for internal API calls

sjkwak:
This change is to expose a knob which allows a user to change timeout for internal API calls and use the Operation Timeout property in the config to specify timeout for all API calls.


core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -368,6 +374,12 @@ final class EventHubsConf private (private val connectionStr: String)
     self.get(OperationTimeoutKey) map (str => Duration.parse(str))
   }
 
+  /** Internal operation timeout. */
+  private[spark] def internalOperationTimeout: FiniteDuration = {
```
sabeegrewal: can we make this a `lazy val` instead of a `def`?

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -368,6 +374,12 @@ final class EventHubsConf private (private val connectionStr: String)
     self.get(OperationTimeoutKey) map (str => Duration.parse(str))
   }
 
+  /** Internal operation timeout. */
```
sabeegrewal: Maybe we can explain here that this is _exactly_ the same as `operationTimeout`. The only difference is in the type (one is a Java `Duration` and the other is a Scala `FiniteDuration`)

sjkwak: good suggestion. will fix it.

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
sjkwak: sure, will do.

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -368,6 +374,13 @@ final class EventHubsConf private (private val connectionStr: String)
     self.get(OperationTimeoutKey) map (str => Duration.parse(str))
   }
 
+  /** it is the same as operationTimeout.
```
basilhariri: Nit - wording:

/** This is the exact same thing as operationTimeout, the only difference is in the types
  * (Java Duration vs Scala FiniteDuration)
  */

sjkwak: Thanks, fixed.

PR #478, opened 2020-02-25
Add support for different partitioning strategies to deal with reader/executor skew

DanMMSFT:
Adds support for alternate preferred partition location strategy.  Defaults back to original hash partition if none is set.

The default hash partitioning can create unbalanced loads for executors, with some executors having 2-3x the number of event hub readers (and AMQP connections/buffers) than others.  This can create OOM issues or lagging/partitioning issues due to the imbalance on partitions (in particular Databricks which doesn't allow fine grained control of executor size outside of VM size)). This is particularly an issue for heavy loads (e.g. near dedicated EH style loads), or cross-region reading of EH data, where it can be difficult to keep up without over provisioning the resources by 2-3x more than should be necessary.  This adds an additional option, "BalancedHash", that attempts to split the load evenly for event hubs, which can help deal with these high load situations in a more efficient manner.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -21,6 +21,7 @@ import java.time.Duration
 import java.util.concurrent.ConcurrentHashMap
 
 import org.apache.http.annotation.Experimental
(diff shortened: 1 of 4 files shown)


sjkwak:
unit test run - https://travis-ci.org/Azure/azure-event-hubs-spark/builds/659388764

PR #486, opened 2020-04-02
Don't close a receiver for epoch receiver case

sjkwak:
The creation of an epoch receiver will disconnect existing receivers and so there is no need to call Close() API on the existing receivers.


sjkwak:
CI run - https://travis-ci.org/github/Azure/azure-event-hubs-spark/jobs/669941058

Issue #488, opened 2020-04-02
Wrong seqNo is set when reading from eventhubs


k4jiang:
Bug Report:

- Actual behavior
Same issue as https://github.com/Azure/azure-event-hubs-spark/issues/462.
The consequence is incredibly severe. I am unable to restart the same Spark Structured Stream without creating a new checkpoint.

The following is the stacktrace:
```
Job aborted due to stage failure: Task 30 in stage 2348.0 failed 4 times, most recent failure: Lost task 30.3 in stage 2348.0 (TID 4058, 10.139.64.6, executor 0): java.lang.IllegalStateException: In partition 30 of http-access-log, with consumer group $Default, request seqNo 19609525 is less than the received seqNo 19684911. The earliest seqNo is 19684804 and the last seqNo is 20231767
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver.checkCursor(CachedEventHubsReceiver.scala:189)
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver.org$apache$spark$eventhubs$client$CachedEventHubsReceiver$$receive(CachedEventHubsReceiver.scala:213)
	at org.apache.spark.eventhubs.client.CachedEventHubsReceiver$.receive(CachedEventHubsReceiver.scala:288)
	at org.apache.spark.eventhubs.rdd.EventHubsRDD.compute(EventHubsRDD.scala:120)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:353)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:317)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:60)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:353)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:317)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:60)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:353)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:317)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:60)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:353)
[...]

PR #490, opened 2020-04-03
Fix IllegalStateException issue and recreate a receiver if the receiver was closed

sjkwak:
The change contains two fixes.
a) fixes IllegalStateException which can happen when committed sequence number in the checkpoint was already expired because of retention policy - #488 
b) recreates a partition receiver if the receive was closed.

### DIFF
--- a/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSource.scala
+++ b/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSource.scala
@@ -160,16 +160,37 @@ private[spark] class EventHubsSource private[eventhubs] (sqlContext: SQLContext,
     val seqNos: Map[NameAndPartition, SequenceNumber] = maxOffsetsPerTrigger match {
       case None =>
         latest
-      case Some(limit) if currentSeqNos.isEmpty =>
-        rateLimit(limit, initialPartitionSeqNos, latest, earliestSeqNos.get)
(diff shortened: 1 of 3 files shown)


sjkwak:
CI run link - https://travis-ci.org/github/Azure/azure-event-hubs-spark/builds/670652922

sjkwak:
CI run with an update - https://travis-ci.org/github/Azure/azure-event-hubs-spark/builds/670736673

core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSource.scala
```diff
@@ -160,18 +160,34 @@ private[spark] class EventHubsSource private[eventhubs] (sqlContext: SQLContext,
...
+    logInfo(s"GetOffset: ${seqNos.toSeq.map(_.toString).sorted}")
 
     Some(EventHubsSourceOffset(seqNos))
   }
 
+  private def adjustStartingOffset(
+      from: Map[NameAndPartition, SequenceNumber]): Map[NameAndPartition, SequenceNumber] = {
+    from.map {
+      case (nAndP, seqNo) =>
+        if (seqNo < earliestSeqNos.get(nAndP)) {
```
nyaghma, line 183: In case the starting offset is adjusted, shouldn’t we report data loss or log this change in the starting offset?

PR #492, opened 2020-04-05
Use receiver timeout for underlying API calls to Event Hubs so that connector can retry on failures

sjkwak:
Current retry logic doesn't retry on a failure because underlying API calls to the Event Hubs service uses operation timeout which is also used by the connector. So, when timeout error occurs while accessing the event hub service, there will be no remaining timeout for the connector to retry.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
@@ -51,7 +51,7 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
         s"No clients left to borrow. EventHub name: ${ehConf.name}, " +
           s"ConsumerGroup name: $consumerGroup. Creating client ${count.incrementAndGet()}")
       val connStr = ConnectionStringBuilder(ehConf.connectionString)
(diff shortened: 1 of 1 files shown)


sjkwak:
CI link - https://travis-ci.org/github/Azure/azure-event-hubs-spark/builds/671186274

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -51,7 +51,7 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
         s"No clients left to borrow. EventHub name: ${ehConf.name}, " +
           s"ConsumerGroup name: $consumerGroup. Creating client ${count.incrementAndGet()}")
       val connStr = ConnectionStringBuilder(ehConf.connectionString)
-      connStr.setOperationTimeout(ehConf.operationTimeout.getOrElse(DefaultOperationTimeout))
+      connStr.setOperationTimeout(ehConf.receiverTimeout.getOrElse(DefaultOperationTimeout))
```
nyaghma, line 54: Shouldn’t we use DefaultReceiverTimeout instead of DefaultOperationTimeout?

PR #493, opened 2020-04-06
Couple of fixes related to configuration and improve logging.

sjkwak:
The change contains the following changes.
1) fix an issue of using default operation timeout if receiver timeout is not specified.
2) add a log for data loss in case starting offset is readjusted due to data expiration.
3) validate receiver timeout and throws if the value is greater than operation timeout.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
@@ -355,6 +355,10 @@ final class EventHubsConf private (private val connectionStr: String)
    * @return the updated [[EventHubsConf]] instance
    */
   def setReceiverTimeout(d: Duration): EventHubsConf = {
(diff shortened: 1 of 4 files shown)


sjkwak:
CI run - https://travis-ci.org/github/Azure/azure-event-hubs-spark/builds/671486317

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -355,6 +355,10 @@ final class EventHubsConf private (private val connectionStr: String)
    * @return the updated [[EventHubsConf]] instance
    */
   def setReceiverTimeout(d: Duration): EventHubsConf = {
+    if (d.toMillis > operationTimeout.getOrElse(DefaultOperationTimeout).toMillis) {
```
nyaghma, line 358: We should add a similar check in the setOperationTimeout as well to cover the case when the user just updates the OperationTimeout and sets it to a value less than the receiver timeout.

PR #494, opened 2020-04-06
Support null values in properties

nyaghma:
This PR adds support for the null values in user-defined properties bags. Also, it updates the java client version to the most recent version (3.1.1).


nyaghma:
CI run: 
https://travis-ci.org/github/Azure/azure-event-hubs-spark/builds/671753843

Issue #511, opened 2020-05-28
Application properties of type String are deserialized with extra double quotes


neilhuiz:
Hi there, it appears that the EventHub client is not doing the right thing when deserializing the application properties map (i.e. `properties` column in the RDD/DF). When a string application property is passed from the sender, it is deserialized with extra quotes wrapped around it. As an example, if I send an event with the application property `MyProperty` with a value of `MyValue`, in my Scala code, I get back a property map with `(MyProperty -> "MyProperty")`:

```
+----------+---------+
|        _1|       _2|
+----------+---------+
|MyProperty|"MyValue"|
+----------+---------+
```

The expected behavior would be without the quotes:

```
+----------+-------+
|        _1|     _2|
+----------+-------+
|MyProperty|MyValue|
+----------+-------+
```

My sender is a C# application and is relying on the default serialization behavior when calling `EventData.Properties.Add("MyProperty", "MyValue")`. Other types (e.g. `bool`, `int`) don't exhibit this behavior, though all are turned into strings on the reader's side. If I may speculate, it appears to be related to the use of `org.json4s.jackson.Serialization.write()` in [EventHubsSourceProvider](https://github.com/Azure/azure-event-hubs-spark/blob/master/core/src/main/scala/org/apache/spark/sql/eventhubs/EventHubsSourceProvider.scala) (see ~line 192). Annoyingly, `org.json4s.jackson.Serialization.write("test") == "\"test\""`.

I verified this behavior under both 2.3.13 and 2.3.15 versions of the `azure-eventhubs-spark` package. I also verified under both 2.11.12 and 2.11.8 versions of Scala. Here is Scala code that reproduces the behavior:

```
val connectionString = ConnectionStringBuilder("{secret sauce}")
    .build

val ehConf = EventHubsConf(connectionString)
    .setConsumerGroup("warm")
    .setStartingPosition(EventPosition.fromEnqueuedTime(Instant.now))

val events = spark
    .readStream
    .format("eventhubs")
    .options(ehConf.toMap)
    .load()
    .toDF()

val query = events
    .flatMap(r => r.getAs[Map[String, String]]("properties"))
    .writeStream
    .format("console")
    .start()

query.awaitTermination()
```

And the C# code:

```
var hub = EventHubClient.CreateFromConnectionString("{secret sauce}");
var evt = new EventData(Encoding.UTF8.GetBytes(payload.ToString()));
evt.Properties.Add("MyProperty", "MyValue");
[...]

PR #514, opened 2020-06-11
Stopped serializing String property values

neilhuiz:
As noted in #511, the current implementation re-serializes strings and adds `"` around the values. For example, if your application property value is `My string value`, it becomes `"My string value"` after the values are serialized as JSON during `EventHubsSourceProvider.toInternalRow()`. Non-string values like `Deimal128` and `UnsignedLong` are **not** wrapped in quotes so this makes it extremely difficult for users of the library to effectively distinguish. If the value was actually a string representation of, for example a numeric value, then it would get wrapped while binary representations would not. Finally, if the string really is wrapped in quotes for some reason, then you get double-double quoting: `""Hello quotes!""`.


nyaghma:
Since this PR did not pass the required Travis CI build, I opened another PR https://github.com/Azure/azure-event-hubs-spark/pull/516 to fix the issue noted in https://github.com/Azure/azure-event-hubs-spark/issues/511. Please let me know if anything needs to be updated.

PR #523, opened 2020-07-23
Do not retry receiver close, log and ignore transient errors.

JamesBirdsall:
PartitionReceiver.close() is not a retryable operation: after the first call, it returns the same CompletableFuture. So don't retry it, and just log and ignore any transient failures.


core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -128,7 +129,24 @@ private[client] class CachedEventHubsReceiver private (ehConf: EventHubsConf,
...
+          logInfo(s"(TID $taskId) transient failure closing a receiver, ignoring")
+          Future.successful(null);
+      case t: Throwable =>
+        t.getCause match {
+          case eh: EventHubException if eh.getIsTransient =>
+            logInfo(s"(TID $taskId) transient failure closing a receiver, ignoring")
+            Future.successful(null);
+          case _ =>
+            logInfo(s"(TID $taskId) nontransient failure closing a receiver")
+            throw t
```
brjeffrey, line 147: This would kill the running task.  When a retry is initiated and lands on the same executor would the receiver be marked to not be re-used?  Would we need to remove this from the cache?

JamesBirdsall: The happy path is this:
1) Spark connector gets receiver for a partition out of cache
2) Receiver is not suitable, so connector tries to close it
3) Close succeeds
4) Connector creates a new receiver
5) Spark job continues OK

The sequence of events that we were seeing in your issue is:
3) Close fails with transient error, connector goes into retry loop that accomplishes nothing and eventually throws
4) Spark job crashes

What we want instead is:
3) Close fails with transient error, connector logs the error but otherwise treats it as success
4) Connector continues on happy path and creates a new receiver
5) Spark job continues OK

PR #535, opened 2020-09-01
add AAD auth client to connect ehs

duhuan:
add aad auth support to connect to ehs, more details:
https://docs.microsoft.com/en-us/azure/event-hubs/authenticate-application


core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -561,6 +568,98 @@ final class EventHubsConf private (private val connectionStr: String)
             s"${PartitionPreferredLocationStrategy.values.mkString(",")}"))
   }
 
+  /**
+   * Set whether to use AAD auth to connect eventhubs.
+   * More info about this: https://docs.microsoft.com/en-us/azure/event-hubs/authorize-access-azure-active-directory
+   * There're 2 options for AadAuth: 1) Certificate; 2) Secret, you'll also need to set
```
sjkwak, line 574: Instead of asking users to provide certificate or secret as an input, what about allowing users to provide the implementation of AzureActiveDirectoryTokenProvider.AuthenticationCallback API so that the library can remain flexible in terms of supporting various ways of obtaining an access token?

sjkwak (changes requested):

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
duhuan: good idea!  we can do that, but on the other side, according to azure AD's doc, for app/sp, it only supports either secret or cert to authenticate: https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app.
All ehs configuration parameters are String, if we want to keep consistent with others and use an interface implementation as a parameter, we may have to use reflection to create/call the method in runtime. I can add one more option in "AadAuth" to do that if you prefer, but I highly suspect whether it'll be really used in the feature. :)

sjkwak: Hi duhuan - thank you for your reply. There is a similar implementation done for metric plug-in. So, one option could be to expose a configuration which a user can set for their implementation of an interface that contains the callback API (https://docs.microsoft.com/en-us/java/api/com.microsoft.azure.servicebus.security.azureactivedirectorytokenprovider.authenticationcallback?view=azure-java-stable).

duhuan: yeah, just added the callback support. please see the latest change :)

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -561,6 +570,108 @@ final class EventHubsConf private (private val connectionStr: String)
             s"${PartitionPreferredLocationStrategy.values.mkString(",")}"))
   }
 
+  /**
+   * Set whether to use AAD auth to connect eventhubs.
+   * More info about this: https://docs.microsoft.com/en-us/azure/event-hubs/authorize-access-azure-active-directory
+   * There're 3 options for AadAuth: 1) Certificate; 2) Secret; 3) Callback. You'll also need to set
```
nyaghma, line 576: Hi @duhuan , SJ and I discussed the options and we think we should just keep the Callback option here. In this way, the library is flexible to support various ways of obtaining an access token and users can provide their implementations to use either Certificate or Secret. 
We would like to have your AuthByCertCallBack and AuthBySecretCallBack implementations as samples that users can utilize in order to use the Callback option. It would be great if you could create a new "examples" directory under the root directory and move those implementations there as two examples of how the Callback option can be used for either Secret or Cert authentication.

nyaghma (changes requested):

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
duhuan: yeah, changed. please review the latest iteration.

nyaghma (changes requested):

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -561,6 +566,44 @@ final class EventHubsConf private (private val connectionStr: String)
...
+  }
+
+  def aadAuthCallback(): Option[AadAuthenticationCallback] = {
+    self.get(AadAuthCallbackKey) map (className => {
+      Class.forName(className).newInstance().asInstanceOf[AadAuthenticationCallback]
+    })
+  }
+
+  /**
+   * Set the tenant id if [[AadAuthKey]] is Certificate or Password
```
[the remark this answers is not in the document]
duhuan: Both the callback class and the tenantID is set in EventHubClient.createWithAzureActiveDirectory. I'd still prefer to keep it here as:
1) from logic, they're the same level instead of one from another. If someone has experience of writing AAD auth implementation, they would expect to set the tenant id in the out layer instead of the callback. :)
2) the "authority" parameter in the callback function(acquireToken) actually is a right value, we should not override or change it there, the right way to do is to get tenant id from authority.
3) I still prefer to keep trait clean and do not include any data in it.

core/src/main/scala/org/apache/spark/eventhubs/package.scala
```diff
@@ -33,6 +33,7 @@ package object eventhubs {
 
   val StartOfStream: String = "-1"
   val EndOfStream: String = "@latest"
+  val DefaultEmptyString: String = ""
```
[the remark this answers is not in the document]
duhuan: good catch! Thanks!

nyaghma (approved):
Looks good to me :)

pom.xml
```diff
@@ -170,6 +170,11 @@
             <scope>test</scope>
             <version>3.0.3</version>
         </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>msal4j</artifactId>
+            <version>0.5.0-preview</version>
```
sjkwak, line 176: This is very old version. Let's use the latest one.

core/src/main/scala/org/apache/spark/eventhubs/client/package.scala
```diff
@@ -0,0 +1,7 @@
+package org.apache.spark.eventhubs
+
+package object client {
+  implicit def toJavaFunction[A, B](f: Function1[A, B]) = new java.util.function.Function[A, B] {
```
sjkwak, line 4: Do we need this? I don't see a usage of the API.

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -561,6 +565,31 @@ final class EventHubsConf private (private val connectionStr: String)
...
+   */
+  private def setUseAadAuth(b: Boolean): EventHubsConf = {
+    set(UseAadAuthKey, b)
+  }
+
+  def useAadAuth: Boolean = {
+    self.get(UseAadAuthKey).getOrElse(DefaultUseAadAuth).toBoolean
+  }
+
+  def setAadAuthCallback(callback: AadAuthenticationCallback): EventHubsConf = {
```
sjkwak, line 582: Can you please add API summary for new APIs?

sjkwak (changes requested):

core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
duhuan: sure, done.

core/src/main/scala/org/apache/spark/eventhubs/client/package.scala
duhuan: removed

pom.xml
duhuan: changed to the latest 1.7.0

nyaghma (approved):

nyaghma (approved):

pom.xml
```diff
@@ -170,6 +170,11 @@
             <scope>test</scope>
             <version>3.0.3</version>
         </dependency>
+        <dependency>
```
sjkwak, line 173: The library doesn't need MSAL dependency and it is required by user's application when implementing the OAuth callback, right?

sjkwak (approved):

PR #544, opened 2020-10-14
Dynamic added partitions

nyaghma:
This PR supports added partitions dynamically without requiring a job restart. It adds a new config knob (DynamicPartitionDiscovery) to EventHubsConfig. If DynamicPartitionDiscovery is enabled,  the client reads the partition count from runtimeInfo periodically (every 5 minutes) in order to update it in case a new partition has been added.


core/src/main/scala/org/apache/spark/eventhubs/EventHubsConf.scala
```diff
@@ -496,10 +501,29 @@ final class EventHubsConf private (private val connectionStr: String)
     self.get(SlowPartitionAdjustmentKey).getOrElse(DefaultSlowPartitionAdjustment).toBoolean
   }
 
+  /**
+   * Set the flag for dynamic partition discovery. If dynamic partition discovery is disabled
```
nyaghma, line 505: update the description

core/src/main/scala/org/apache/spark/eventhubs/client/EventHubsClient.scala
```diff
@@ -164,15 +165,41 @@ private[spark] class EventHubsClient(private val ehConf: EventHubsConf)
    *
    * @return partition count
    */
-  override lazy val partitionCount: Int = {
+  override def partitionCount: Int = {
```
nyaghma, line 168: Add try and catch

sjkwak (approved):

PR #552, opened 2020-10-29
Retry client create

nyaghma:
The library calls EventHubClient.createFromConnectionStringSync(…) in a loop until the result is not null, but it doesn’t catch exceptions that the method may throw (IOException and EventHubException). Therefore, in those situations, it doesn’t retry properly. This PR adds a try-catch block to make sure the client creation is being retried when there is a transient error from the underlying client.


JamesBirdsall (approved):

core/src/main/scala/org/apache/spark/eventhubs/client/ClientConnectionPool.scala
```diff
@@ -56,28 +57,46 @@ private class ClientConnectionPool(val ehConf: EventHubsConf) extends Logging {
...
+              s"Ignoring transient failure ${ehException} from EventHubClient.createFromConnectionStringSync for EventHub name: ${ehConf.name}")
+            client = null
+          case t: Throwable =>
+            t.getCause match {
+              case ehException: EventHubException if ehException.getIsTransient =>
+                logInfo(
+                  s"Ignoring transient failure ${ehException} from EventHubClient.createFromConnectionStringSync for EventHub name: ${ehConf.name}")
+                client = null
+              case _ =>
+                logInfo(
```
sjkwak, line 96: Use RetryJava utility method which handles transient failures.

sjkwak (changes requested):

sjkwak (approved):

PR #579, opened 2021-01-19
Fix typo in docs when importing datetime library

emavgl:
Hi there,

I changed a line in the pyspark documentation.
The existing line `import datetime from datetime as dt` returns `SyntaxError: invalid syntax`.
The correct importing statement is instead: `from datetime import datetime as dt`.


abij:
Facing same 'issue', can you add `import json`, so the example can actually works?

emavgl:
Done.

sjkwak (approved):

PR #590, opened 2021-04-09
Service principal aad auth - params added to service principal callback

kklimexk:
Problem I had here is that service principal callback couldn't have any parameters passed to it:
```scala
case class AuthBySecretCallBack() extends AadAuthenticationCallback {

  implicit def toJavaFunction[A, B](f: Function1[A, B]) = new java.util.function.Function[A, B] {
    override def apply(a: A): B = f(a)
  }
  override def authority: String = "your-tenant-id"

  val clientId: String = "your-client-id"
  val clientSecret: String = "your-client-secret"

//...
}
```
In the above code authority, clientId, clientSecret are hard coded values which is unacceptable in my case. I tried to do sth like:
```scala
  override def authority: String = dbutils.secrets.get(SecretScopeName, TenantID)
  val clientId: String = dbutils.secrets.get(SecretScopeName, ClientId)
  val clientSecret: String = dbutils.secrets.get(SecretScopeName, ClientSecret)
```
but then I got an error. So in this pull request I changed the way of defining this service principal auth callback and I added parameters to it, now it looks like this:
```scala
class AuthBySecretCallBack(params: String*) extends AadAuthenticationCallback(params) {

  implicit def toJavaFunction[A, B](f: Function1[A, B]) = new java.util.function.Function[A, B] {
    override def apply(a: A): B = f(a)
  }
  override def authority: String = "your-tenant-id" //or taken from "params"

  val clientId: String = "your-client-id" //or taken from "params"
  val clientSecret: String = "your-client-secret" //or taken from "params"

  override def acquireToken(audience: String, authority: String, state: Any): CompletableFuture[String] = try {
    var app = ConfidentialClientApplication
      .builder("clientId", ClientCredentialFactory.createFromSecret(this.clientSecret))
      .authority("https://login.microsoftonline.com/" + authority)
      .build

    val parameters = ClientCredentialParameters.builder(Collections.singleton(audience + ".default")).build

    app.acquireToken(parameters).thenApply((result: IAuthenticationResult) => result.accessToken())
  } catch {
    case e: Exception =>
      val failed = new CompletableFuture[String]
      failed.completeExceptionally(e)
      failed
  }
}
```
Thanks to "params" argument I can now pass parameters to callback and run my spark job without any issues.

PR #596, opened 2021-05-19
Add parameters to Aad Auth callback class

nyaghma:
This PR adds support for passing a sequence of strings as a parameter to the aad auth callback class and updates the example on how to use it with secret scope.


kklimexk:
@nyaghma This looks fine, but what is the difference between your code and my pull request (https://github.com/Azure/azure-event-hubs-spark/pull/590)? I think idea is the same.

sjkwak (approved):

PR #614, opened 2021-09-28
add documentation for using aad auth in pyspark

nyaghma:
This PR adds a document (use-aad-authentication-pyspark.md) that shows how to use AAD authentication in Pyspark.


j2k850:
@nyaghma I have a usecase where I would like to use AAD auth in Pyspark. I see this PR is still requires a review before it can be merged into master. Regardless, are the instructions in good enough shape for me to use.

PR #615, opened 2021-10-01
providing to_avro and from_avro functionalities

nyaghma:
This PR adds a plugin that provides support for schema registry in spark. The plugin creates a separate jar file and includes to_avro and from_avro functions for serializing and deserializing payloads with respect to a schema in the schema registry.
The PR includes a doc with both producer and consumer examples.


scaladevspark:
any change to get this PR merged? is there any other way to work with Azure Schema Registry via Spark Structured Streaming?

Issue #612, opened 2021-08-31
java.util.concurrent.CompletionException: java.util.concurrent.RejectedExecutionException: ReactorDispatcher instance is closed


AmanPandey123:
Hi , 

We have pyspark code fetching data from single consumer group with multiple partitions from different event Hub instances under same Event Hub namespace . The jobs run 24*7 without any downtime expected . iterating in loop for every 15 minutes and pulling data and writing to ADLS Gen 1 Sink. 
recently after upgrading to 

Databricks Runtime 8.3 , Scala 2.12 , Spark 3.1.1  and 
event hub library version -  azure-eventhubs-spark_2.12:2.3.20
 the jobs ran fine without any issues for the first 8-9 days , but now we are seeing failed instances with error as below
 
An error occurred while calling o2180548.count. : org.apache.spark.SparkException: Job aborted due to stage failure: Task 11 in stage 157672.0 failed 4 times, most recent failure: Lost task 11.3 in stage 157672.0 (TID 1205598) (10.139.64.6 executor 3312): java.util.concurrent.CompletionException: java.util.concurrent.RejectedExecutionException: ReactorDispatcher instance is closed. at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292) at java.util.concurrent.Com Command Line: 72

After restarting the cluster the issue was gone . I am afraid , i might encounter the issue again . What is the cause for this issue ?

jdbree:
We have the same issue with our pyspark code, see my comment to issue #363. 

Databricks Runtime 8.4
Library versions:
- com.microsoft.azure:azure-eventhubs-spark_2.12:2.3.21
- com.microsoft.azure:azure-eventhubs:3.3.0

The error message:

org.apache.spark.SparkException: Job aborted due to stage failure: Task 1 in stage 3350.0 failed 1 times, most recent failure: Lost task 1.1 in stage 3350.0 (TID 7401) (10.139.64.4 executor driver): java.util.concurrent.CompletionException: java.util.concurrent.RejectedExecutionException: ReactorDispatcher instance is closed.

and further down:

Caused by: java.util.concurrent.CompletionException: java.util.concurrent.RejectedExecutionException: ReactorDispatcher instance is closed.
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
...

[...]

PR #620, opened 2021-11-18
recreate the client if ReactorDispatcher is closed

nyaghma:
Spark Task Preemption could result in closing the ReactorDispatcher object in the client object if it happens during a blocking IO operation on a receiver/sender channel. This makes the client unusable and any upcoming task using that client gets java.util.concurrent.RejectedExecutionException with "ReactorDispatcher instance is closed" message and fails the entire job. (issue #612)
This PR adds exception handling in the receive method in the CachedEventHubsReceiver object to take care of this situation. In this case, the client should be recreated and the new client should be used to create an EventHubsReceiver and update the cache.

### DIFF
--- a/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
+++ b/core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
@@ -391,11 +391,11 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
     } catch {
       case completionExecution: CompletionException =>
         val exceptionCause = completionExecution.getCause
-        if (exceptionCause != null &&  exceptionCause.isInstanceOf[RejectedExecutionException] && exceptionCause.getMessage.contains("ReactorDispatcher")) {
+        if (exceptionCause != null &&  exceptionCause.isInstanceOf[RejectedExecutionException] && exceptionCause.getMessage.contains("ReactorDispatcher instance is closed")) {
           // reactor dispatcher closed case
           logInfo(s"(TID $taskId) EventHubsCachedReceiver receive execution for namespaceUri ${ehConf.namespaceUri} " +
             s"EventHubNameAndPartition $nAndP consumer group ${ehConf.consumerGroup.getOrElse(DefaultConsumerGroup)} " +
-            s"failed with $completionExecution. Try to recerate the entire CachedEventHubsReceiver instance in order to " +
+            s"failed with $completionExecution. Try to recreate the entire CachedEventHubsReceiver instance in order to " +
             s"use a fresh EventHubClient from the underlying java SDK, then try receiving events again.")
           receiver.client.close();
           receiver = CachedEventHubsReceiver(ehConf, nAndP, requestSeqNo)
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -386,7 +386,35 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
...
+    try {
+      receiver.receive(requestSeqNo, batchSize)
+    } catch {
+      case completionExecution: CompletionException =>
+        val exceptionCause = completionExecution.getCause
+        if (exceptionCause != null &&  exceptionCause.isInstanceOf[RejectedExecutionException] && exceptionCause.getMessage.contains("ReactorDispatcher")) {
+          // reactor dispatcher closed case
+          logInfo(s"(TID $taskId) EventHubsCachedReceiver receive execution for namespaceUri ${ehConf.namespaceUri} " +
+            s"EventHubNameAndPartition $nAndP consumer group ${ehConf.consumerGroup.getOrElse(DefaultConsumerGroup)} " +
+            s"failed with $completionExecution. Try to recerate the entire CachedEventHubsReceiver instance in order to " +
```
JamesBirdsall, line 398: recreate

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -386,7 +386,35 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
...
+          receiver.client.close();
+          receiver = CachedEventHubsReceiver(ehConf, nAndP, requestSeqNo)
+          receivers.synchronized {
+            receivers.update(key(ehConf, nAndP), receiver)
+          }
+          receiver.receive(requestSeqNo, batchSize)
+        } else if (exceptionCause != null &&  exceptionCause.isInstanceOf[ReceiverDisconnectedException]) {
+          logInfo(s"(TID $taskId) EventHubsCachedReceiver receive execution for namespaceUri ${ehConf.namespaceUri} " +
+            s"EventHubNameAndPartition $nAndP consumer group ${ehConf.consumerGroup.getOrElse(DefaultConsumerGroup)} " +
+            s"failed becuase another receiver for the same <NS-EH-CG-Part> combo has been created and cause this one " +
```
JamesBirdsall, line 409: because, caused

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -386,7 +386,35 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
...
+            s"to get discnnected. The full error is: $completionExecution. Throw the exception so that the driver can " +
```
JamesBirdsall, line 410: disconnected

core/src/main/scala/org/apache/spark/eventhubs/client/CachedEventHubsReceiver.scala
```diff
@@ -386,7 +386,35 @@ private[spark] object CachedEventHubsReceiver extends CachedReceiver with Loggin
...
+          logInfo(s"(TID $taskId) EventHubsCachedReceiver receive execution for namespaceUri ${ehConf.namespaceUri} " +
+            s"EventHubNameAndPartition $nAndP consumer group ${ehConf.consumerGroup.getOrElse(DefaultConsumerGroup)} " +
+            s"failed with $completionExecution. Try to recerate the entire CachedEventHubsReceiver instance in order to " +
+            s"use a fresh EventHubClient from the underlying java SDK, then try receiving events again.")
+          receiver.client.close();
+          receiver = CachedEventHubsReceiver(ehConf, nAndP, requestSeqNo)
+          receivers.synchronized {
+            receivers.update(key(ehConf, nAndP), receiver)
+          }
+          receiver.receive(requestSeqNo, batchSize)
```
JamesBirdsall, line 405: What if this attempt also throws? Maybe this change should restructured as a do-while or similar: the loop exits because of a successful receive, the loop exits because we throw out, or we handle the exception and loop so the retry has all the same protections as the first attempt. Might want a retry count and exit on too many retries also so that we don't loop forever if something is bad.

nyaghma: This handler just takes care of the case where the ReactorDispatcher has been closed and the client is not usable anymore. If recreating the client doesn't resolve the issue then probably a more serious issue is going on and it's better to fail the job.  Also, this client recreation would be tried up to 4 times by spark, since a failure results in failing the task and spark retries tasks 4 times before failing a job. Therefore, I'm not sure if a do-while would be necessary here.

JamesBirdsall: OK, so effectively we already have 4x retry on this, just at a different level. Good enough.

hmlam (approved):

PR #650, opened 2022-08-18
Schema Registry Receiver

yamin-msft:
Thanks for contributing! We appreciate it :) 

For a Pull Request to be accepted, you must:
- Run scalafmt on your code using the `.scalafmt.conf` present in this project
- All tests must pass when you run `mvn clean test` 


schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,104 @@
...
+      resultRow
+
+    case _ =>
+      null
+  }
+
+  override def nullSafeEval(input: Any): Any = {
+      val message = new com.azure.core.models.MessageContent()
+        .setBodyAsBinaryData(BinaryData.fromBytes(input.asInstanceOf[Array[Byte]]))
+        .setContentType(("avro/binary+"+schemaRegistryConstructor.schemaId))
```
hmlam, line 77: Does it make sense for us to abstract this out and have schemaRegistryConstructor to return the full contentType instead? schema registry technically should know the type based on schema group (SchemaGroup actually suppose to have schemaType defined).

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/SchemaRegistryExceptions.scala
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.microsoft.azure.schemaregistry.spark.avro
+
+final case class MissingPropertyException(private val message: String) extends Exception(message) {}
+
+final case class SchemaNotFoundException(private val message: String) extends Exception(message) {}
```
hmlam, line 22: I might be missing it but I don't seem to see where we use this?

schemaregistry-avro/docs/PySpark/schema-registry-example-pyspark.md
```diff
@@ -0,0 +1,100 @@
...
+   * Deserialize the data using `from_avro` function.
+
+### Wrapper `from_avro` Function for PySpark
+First, we need to define a wrapper function in order to access and use the `com.microsoft.azure.schemaregistry.spark.avro.functions.from_avro` 
+from JVM properly.
+
+```python
+from pyspark.sql.column import Column, _to_java_column
+
+def from_avro(col, propMap, configMap):
```
hmlam, line 47: This doesn't seem right based on the actual definition of from_avro you defined... can you update the sample for PySpark to reflect the usage correctly?

schemaregistry-avro/docs/schema-registry-example.md
```diff
@@ -0,0 +1,96 @@
...
+  configs.put(SCHEMA_REGISTRY_CLIENT_SECRET_KEY, schemaRegistryClientSecret)
+```
+
+#### Deserialize Data
+```scala
+import com.microsoft.azure.schemaregistry.spark.avro.functions._
+import com.microsoft.azure.schemaregistry.spark.avro.SchemaGUID
+
+val schemaGUIDString = "<YOUR_SCHEMA_GUID>"
+val parsed_df = df.select(from_avro($"body", SchemaGUID(schemaGUIDString), props) as "jsondata")
```
hmlam, line 89: where is "props" defined?

hmlam (changes requested):

schemaregistry-avro/pom.xml
```diff
@@ -0,0 +1,216 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.36</version>
+    </dependency>
+    <dependency>
+      <groupId>com.azure</groupId>
+      <artifactId>azure-core</artifactId>
+      <version>1.28.0</version>
```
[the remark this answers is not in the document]
yamin-msft: updated

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,104 @@
...
+      val genericRecord = schemaRegistryConstructor.serializer.deserialize(message, TypeReference.createInstance(classOf[GenericRecord]))
+
+    try {
+      avroConverter.deserialize(genericRecord)
+  } catch {
+      case NonFatal(e) => parseMode match {
+        case PermissiveMode => //nullResultRow
```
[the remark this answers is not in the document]
yamin-msft: return nullresultrow

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/SchemaRegistryExceptions.scala
yamin-msft: you are right that SchemaNotFoundException is not in use

schemaregistry-avro/docs/PySpark/schema-registry-example-pyspark.md
yamin-msft: modfified

schemaregistry-avro/docs/schema-registry-example.md
yamin-msft: modified

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: Will include this in the next update. Created a workitem for it.

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,78 @@
...
+  @transient private lazy val avroConverter = {
+    new AvroDeserializer(schemaRegistryConstructor.expectedSchema, dataType)
+  }
+
+  override def nullSafeEval(input: Any): Any = {
+    val message = new com.azure.core.models.MessageContent()
+      .setBodyAsBinaryData(BinaryData.fromBytes(input.asInstanceOf[Array[Byte]]))
+      .setContentType(("avro/binary+"+schemaRegistryConstructor.schemaId))
+    val genericRecord = schemaRegistryConstructor.serializer.deserialize(message, TypeReference.createInstance(classOf[GenericRecord]))
+    logInfo("Finished deserializing Message Content")
```
hmlam, line 54: General comment on logging:
- Can we be more uniform in where we log? Here we are logging after the execution is done but in line 57 you are logging before the line of code is executed. Let's log before the execution to keep things uniform.
- The description should say something along the line of "Deserializing EventData $messageOffset using schema $schemaGuid"
- Can we log the task id? [EventHubUtils](https://github.com/Azure/azure-event-hubs-spark/blob/6efb19211ffed780930d48eb1744ec1863338e01/core/src/main/scala/org/apache/spark/eventhubs/EventHubsUtils.scala#L159) already have an implementation on getting the TaskId.
- This get logged per row in the RDD, so can we be a bit more descriptive so that user get a bit more information? e.g. the generic record being null or not, message id, schema Guid, etc. This helps us debug things when customer capture the logs.

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,78 @@
...
+
+    try {
+      logInfo("Deserializing Avro data to Catalyst data")
+      avroConverter.deserialize(genericRecord)
+    } catch {
+      case e:Throwable =>
```
hmlam, line 60: 1. Do we need to catch Throwable here? Throwable is a superclass that include both Error and Exception, but I don't think I see you handled anything about the Error class for example. It is ok to catch Throwable if we intend to do something about it (e.g. log details) but still, I think we should just throw back whatever we have got in that case.
2. Can we also log error in this case? In case user just ignore all exception the log should always include the error and stack for debugging purposes.

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  override def nullSafeEval(input: Any): Any = {
+    val message = new com.azure.core.models.MessageContent()
+      .setBodyAsBinaryData(BinaryData.fromBytes(input.asInstanceOf[Array[Byte]]))
+      .setContentType(("avro/binary+"+schemaRegistryConstructor.schemaId))
+    val genericRecord = schemaRegistryConstructor.serializer.deserialize(message, TypeReference.createInstance(classOf[GenericRecord]))
+    logInfo("Finished deserializing Message Content")
+
+    try {
```
hmlam, line 56: Should this try/catch block actually include schemaRegistryConstructor.serializer.deserialize as well?

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,78 @@
...
+      logInfo("Deserializing Avro data to Catalyst data")
+      avroConverter.deserialize(genericRecord)
+    } catch {
+      case e:Throwable =>
+      {
+        throw new DeserializationException("Failed to deserialize, error message: \n" + e)
```
hmlam, line 62: Before we commit to this wrapper approach I want to check one thing. What can AvroConverter throws back that user cannot handle without detecting it with DeserializationException? If there is not many exception types to deal with we may be ok not to wrap it with a new exception class.

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
```diff
@@ -0,0 +1,78 @@
...
+      }
+      case e => logInfo(s"Error Message: ${e.getMessage}")
```
hmlam, line 64: Is this needed? If you are catching Throwable, I think that covers everything?

hmlam (changes requested):

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: fixed

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: fixed

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: fixed

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: fixed

schemaregistry-avro/src/main/scala/com/microsoft/azure/schemaregistry/spark/avro/AvroDataToCatalyst.scala
yamin-msft: There are 2 kinds of exceptions from AvroConverter and more from deserialize function. Currently I keep the wrapper for now. If you think unwrapping will work better, please let me know.

hmlam (approved):

hmlam (approved):

